Issue #11, opened 2019-03-04
Basic playlist support


hrkfdn:
Implement loading/saving of playlists. Basic features that we want:

- [x] ~As Spotify supports playlist folders, we need a way to hierarchically navigate through folders (maybe with [cursive_tree_view](https://github.com/BonsaiDen/cursive_tree_view)?)~ ([the API doesn't offer folder information](https://github.com/spotify/web-api/issues/38#issuecomment-396925978))
- [x] Load and display playlists
- [x] Playlists can be queued
- [x] Playlists should only be loaded on start-up
- [x] .. and if explicitly requested by user
- [x] Cache loaded playlists for next startup. Use `snapshot_id` field to check if a cached playlist's tracklist needs to be updated.
- [x] Queues can be saved to playlists
- [x] Playlists can be deleted

The playlist view should be an extra screen (like Search/Queue).

norcalli:
You could take a look at my branch from an older version I made on the day you released it where I implemented a tree view for playlists. I included cursive-tree-view since I made some changes, but you can probably work around it (and also one of them was merged). https://github.com/norcalli/ncspot

barzamin:
Ideally there'd be some way of accessing your "saved songs" library as a playlist, too.

hrkfdn:
@norcalli Some solid ideas in there. The codebase has changed quite a bit, but I might implement some of those ideas.

norcalli:
@barzamin "Saved songs" has a special endpoint, so it is definitely accessible. It's probably best served on a separate screen with one long playlist and the ability to search entries in it. Which reminds me, that I was going to add searching within playlists after pre-caching all of the entries. I have quite a lot of songs, but even still, 3000 entries is not that much space in memory on a modern computer.

norcalli:
@hrkfdn oh yeah one thing that took the longest to implement that I forgot to add was to immediately play a song when you do an "append to queue" action if nothing is currently playing (as in a stopped state vs play/pause).

im-n1:
Not sure if this fits in here but its related to playlists.
I'm missing a feature where I can browse tracks of a playlist. Now I can only queue whole playlist not certain tracks.

hrkfdn:
@im-n1 It's on my list :)

Ads20000:
[...]

PR #13, opened 2019-03-04
Implement basic playlist that simply queues songs

felipesere:
This is a fairly crude implementation of playlists.
It pulls the users playlist on boot and then displays them when hitting F4.
Pressing `return` on a playlist queues the songs.

Stuff that could be cool for either this PR or followups:
* only load the playlist when actually jumping to the view
* allow refreshing the playlists
* disconnect playing of a playlist from the queue and allow to repeat songs etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ have them configurable.
   * `F1`: Debug log
   * `F2`: Queue
   * `F3`: Search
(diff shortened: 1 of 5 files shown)


hrkfdn:
Thanks! We'll probably want a tree view (#11) in the future, but this is great for a start! Good work!

PR #14, opened 2019-03-04
Load playlists only when opening the playlist screen

felipesere:
Make loading the playlist asynchronous by not doing it `main.rs` and scheduling an event.


hrkfdn:
This will load the playlists every time the playlist screen is opened, right? We probably want to do this only once and after that by demand.

hrkfdn:
I've merged this with a few changes. Thank you :)

felipesere:
Oh cool! How would I go about adding callbacks that are only local to a single screen e.g. the playlist screen?

PR #17, opened 2019-03-05
Add main layout view and status bar

KoffeinFlummi:
I added a status bar containing playback status, name and duration with a progress bar spanning the entire line above. I'm not sure yet whether that's a good use of space, or if that should be integrated into the single line (this might look odd because of being off-center). It could potentially automatically be removed if the window height drops below a certain treshold.

To accomplish this, I also made some other changes:
- Wrapping all main views inside a custom layout for view switching
- Changing the load command to use `FullTrack` instead of `SpotifyId`
- Keeping track of the currently playing track in `Spotify`
- Renaming `PlayerState` to `PlayerStatus` to prevent misunderstandings, since the state of the entire player consists of that status, the current track and other data.
- Keeping track of the elapsed time of the current song. This is really hacky, but unless I missed it there's currently no easy way to get this from `librespot`. This should probably be added upstream. As you can see in the gif, my method is not perfect and sometimes goes beyond the total duration.
- Adding getters and setters for the `RwLock`s in `Spotify` to increase readability.

### DIFF
--- a/src/events.rs
+++ b/src/events.rs
@@ -1,13 +1,16 @@
 use crossbeam_channel::{unbounded, Receiver, Sender, TryIter};
 use cursive::{CbFunc, Cursive};
 
+use rspotify::spotify::model::track::FullTrack;
+use spotify::PlayerStatus;
+
 use queue::QueueChange;
-use spotify::PlayerState;
 use ui::playlist::PlaylistEvent;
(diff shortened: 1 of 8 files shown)


hrkfdn:
Great work! Thanks a lot!

PR #25, opened 2019-03-09
Add a vim-style command line

KoffeinFlummi:
As promised, adds a basic vim-style command line. I replaced all the existing (global) keybindings with commands. This allows easy rebinding, although at the moment they remain hardcoded.

Command callbacks return `Result<Option<String>, String>`. If an error is returned, the text is displayed above the command line.

The `:search` command allows entering the query directly, although the `on_submit` callback is currently not called, so you still have to press Enter.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -1,0 +1,49 @@
+use std::collections::HashMap;
+
+use cursive::Cursive;
+
+pub struct CommandManager {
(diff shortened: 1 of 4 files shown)


hrkfdn:
Great work, thanks! I've merged this with a minor change: the search command will join the arguments with a space in between.

PR #26, opened 2019-03-09
turn off writing log to file

andrewhalle:
I tend to use ncspot in whatever project directory I'm currently in, so I have a bunch of these 'ncspot.log' files hanging around now. Turning this off, but leaving the structure there so it can be a command line option later

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -43,7 +43,7 @@ use queue::QueueEvent;
 use spotify::PlayerEvent;
 use ui::playlist::PlaylistEvent;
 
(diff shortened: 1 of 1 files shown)


hrkfdn:
Merged, thanks. This should probably we followed up with a command line flag to re-enable file logging at some point, as the debug log gets lost when the application crashes.

andrewhalle:
If you don’t mind adding another dependency, I hear clap is a pretty good crate for command line args, I can add that in.

Issue #9, opened 2019-03-03
Should work with media keys


andrewhalle:
Official client responds to media keys (play/pause, fast foward, reverse) even when the window doesn't have focus, ideally the same thing could be done here.

Herbstein:
This is IMO a must if the program should be completely useful in daily life

hrkfdn:
Indeed, added to the list.

andrewhalle:
Been looking into this a bit. It seems like hooking into global keyboard events is highly display server/operating system specific, so perhaps this should be a compile-time feature like pulse/port audio. In X11, the function to call is [XGrabKey](https://tronche.com/gui/x/xlib/input/XGrabKey.html). I'm sure something similar exists in Cocoa for mac.

hrkfdn:
On Linux we probably want to catch the media-keys via DBUS or directly go for MPRIS for further desktop integration.

andrewhalle:
That makes sense. Should be optional though to support environments without DBUS (like running directly on the virtual terminal, which currently works)

norcalli:
Yeah, the preferred way to do this is usually to register with DBUS on linux through MPRIS. I currently have my buttons mapped to mpris commands.

im-n1:
I dont think its the best idea to force register keys. Instead of that I would go for some kind of listening for certain commands or even the Dbus.

The thing here is I dont want to start play music when I hit pause media key while watching a movie in smplayer for example.

hrkfdn:
This is taken care of thanks to @KoffeinFlummi, at least on dbus systems.

Bettehem:
I think this is an issue with macOS, not ncspot. You could assign a keyboard shortcut to your desired dbus command using something like [iCanHazShortcut](https://github.com/deseven/icanhazshortcut) to create custom keyboard shortcuts on macOS.

tonglil:
Would appreciate if anyone has found out how to control `ncspot` w/ media keys on macos to share.

I can use BTT to call a command in the terminal, but unsure what that is right now - what needs to be sent with `dbus-send`?

deed02392:
You need to send an MPRIS compatible command over dbus. I made a shell script I called from Raycast script like this:

```sh
#!/bin/bash

# @raycast.schemaVersion 1
# @raycast.title Play/Pause for ncspot
# @raycast.mode silent

# @raycast.icon 🎧

# Documentation:
# @raycast.description Play/Pause for ncspot via dbus
[...]

PR #27, opened 2019-03-11
Implement MPRIS D-Bus spec

KoffeinFlummi:
Implements a D-Bus server responding to the `org.mpris.MediaPlayer2` and `org.mpris.MediaPlayer2.Player` specifications as an optional feature:

https://specifications.freedesktop.org/mpris-spec/latest/

`org.mpris.MediaPlayer2.TrackList` and `org.mpris.MediaPlayer2.Playlists` are not implemented, but this is enough to read metadata and control playback.

This should satisfy #9 on D-Bus systems at least.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,9 +22,12 @@ rspotify = "0.2.5"
 serde = "1.0"
 serde_derive = "1.0"
 toml = "0.4"
+tokio = "0.1.7"
 tokio-core = "0.1"
 tokio-timer = "0.2"
 unicode-width = "0.1.5"
+dbus = { version = "0.6.4", optional = true }
+dbus-tokio = { version = "0.3.0", optional = true }
 
 [dependencies.librespot]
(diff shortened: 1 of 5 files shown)


hrkfdn:
Again, great work. Merged.
Additionally, I fixed a crash that happened because of a track in my playlist that had no album art.

It seems that this does not yet satisfy GNOME for player integration. Do you happen to know what is missing besides MPRIS? I couldn't really find any documentation on the topic. The player is displayed in the notification area and play/next/prev buttons work, but the track metadata is missing.

KoffeinFlummi:
> Additionally, I fixed a crash that happened because of a track in my playlist that had no album art.

Huh. I didn't think that was possible.

> EDIT: Seems that property changes need to be signaled using PropertiesChanged for this.

Yes, I foolishly assumed that dbus-rs would handle that for me, but it seems not. I set up a VM and am looking into it.

PR #28, opened 2019-03-12
Add playlist-shuffle command

norcalli:
I opted not to add a keyboard shortcut for now. The command interface by itself is useful since shuffling is not a common operation. As it stands, the position in the playlist is maintained since it's referenced by index. This means that the next position in the playlist will be continued. I considered resetting the position, but I wasn't certain.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,7 @@ tokio-timer = "0.2"
 unicode-width = "0.1.5"
 dbus = { version = "0.6.4", optional = true }
 dbus-tokio = { version = "0.3.0", optional = true }
+rand = "0.6.5"
 
 [dependencies.librespot]
 git = "https://github.com/librespot-org/librespot.git"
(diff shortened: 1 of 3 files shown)


hrkfdn:
Hey there, thanks!

This shuffles the playlist in-place, right? I am unsure if this is the way to go forward. Maybe it'd be nicer to have it play the next track randomly and preserve the queue order in memory. What do you think?

norcalli:
What is the benefit of preserving the queue order? Additionally, those two aren't mutually exclusive. What you're referring to would be more like "enabling shuffle mode", whereas this is a command that shuffles the playlist. It was inspired directly by the `mpv` command equivalent `playlist-shuffle`.

norcalli:
There is a flaw with this implementation where the song displayed is incorrect once shuffled. I'll fix it if it is determined that you are interested in proceeding with this PR.

hrkfdn:
We definitely want to preserve order in case you load a playlist, add/remove a track and save it again. If the playlists were shuffled in between, I'd assume that would confuse a lot of users. Is there any reason you prefer the in-place shuffle over the shuffle playback mode?

PR #30, opened 2019-03-12
Refactor DBus server, implement PropertiesChanged

KoffeinFlummi:
Refactored the DBus server, yanking out all the `dbus-tokio` stuff since it failed in interesting and annoying ways when I tried to implement a manual poll loop and manually sending out `PropertiesChanged` signals when needed.

I also fixed a typo in the interface name and the artists being strings instead of arrays (this caused an error in the GNOME shell on top of the missing events).


hrkfdn:
Looks great! Thanks!

PR #37, opened 2019-03-17
Add rebindable keys, refactor lists

KoffeinFlummi:
This is a big one; It sort of escalated from the initial problem.

To make keys rebindable, I had to move the list-specific keybindings ({play,queue,delete} selected) out of their views, and since Cursive's own views (`LinearLayout`, `ListView`) hardcode the arrow keys for navigation, I replaced the `LinearLayout`s of `SplitButton`s with a custom `ListView` and a corresponding trait `ListItem`, implemented for `Track` and `Playlist`. This can easily be expanded later for other lists (browsing, library, etc.) and albums/artists.

This view just holds a reference to the actual data and not the data itself. This is a bit cleaner IMO since it doesn't require manually repopulating the views after playlist/queue changes. In conjunction with this I also moved the queue into an internal `RwLock` as was already planned.

Since I had to reimplement navigation, at the moment you cannot switch from the search results list to the form field via arrow keys. Instead you have to use `F1` or `/` and `Esc` to get back out.

I also moved the command definitions out of main.rs and implemented the Playlist/Queue/Search views as actual views or view wrappers instead of using a View struct member from the outside.

Keybinding config format:

```
[keybindings]
"k" = "move up"
"j" = "move down"
"h" = "move left"
"l" = "move right"
"Enter" = "play selected"
```

All of Cursive's `event::Key` options are recognized, in addition to arbitrary `char`s.


hrkfdn:
Great work again, thanks.
I added a small fix to make playlist cache parsing working again. I didn't realize the PlaylistStore implementing the traits required by serde was optional, so the restructuring and removal of PlaylistStore is a welcome change. However, the type here needs to be adjusted to Vec as well. Otherwise deserialization will fail and the cache can't load.

I'd love to merge this, but I think we need to restore arrow navigation first. I have only skimmed the diff, so I don't know yet, what exactly is missing.

src/main.rs
```diff
@@ -206,151 +185,20 @@ fn main() {
...
+    #[cfg(feature = "mpris")]
     {
-        cursive.add_global_callback(Key::F1, move |s| {
-            s.call_on_id("main", |v: &mut ui::layout::Layout| {
-                v.set_view("queue");
-            });
-        });
+        let mpris_manager = mpris_manager.clone();
+        cmd_manager.register_callback(Box::new(move || {
+            mpris_manager.update();
```
hrkfdn: Why is an MPRIS update necessary necessary on command execution? Shouldn't the player events cover the data updates?

hrkfdn:
Okay, I have merged this and implemented the necessary behavior for the arrow keys (only up/down) for now. We may want to consider deleting the binding configurations for the arrow keys?

KoffeinFlummi:
> Okay, I have merged this and implemented the necessary behavior for the arrow keys (only up/down) for now. We may want to consider deleting the binding configurations for the arrow keys?

Sorry for the confusion; the arrow navigation should already work, only the switching between list and the search field via arrow keys doesn't.

src/main.rs
KoffeinFlummi: Looks like it isn't. I implemented this because it was previously run on received commands, but it looks like it can be completely removed. Woops.

hrkfdn: Mhh.. yea, I guess we can remove this. MPRIS seems to be outdated either way.

hrkfdn:
> > Okay, I have merged this and implemented the necessary behavior for the arrow keys (only up/down) for now. We may want to consider deleting the binding configurations for the arrow keys?
> 
> Sorry for the confusion; the arrow navigation should already work, only the switching between list and the search field via arrow keys doesn't.

Odd. I can't browse through playlists without the changes I just committed. The arrow up/down events seem to be intercepted by the ScrollView.

src/main.rs
hrkfdn: Disregard the last sentence, I just updated to GNOME 3.32 and that may be the reason MPRIS doesn't work anymore. Seems fine now + I removed the double update.

PR #39, opened 2019-03-20
Add scroll bar to ListView, remove ScrollView

KoffeinFlummi:
Implements a scroll bar in `ListView` directly, without having to rely on `ScrollView`. Makes things cleaner in the main views, and stops `ScrollView` from consuming arrow key events, making those rebindable.


hrkfdn:
Good stuff!

PR #44, opened 2019-03-24
Credentials helper if config not present or unparseable

### LINKED ISSUE #1: Implement login dialog
Requiring new users to create a configuration themselves is not user-friendly.

Create a dialog asking for Spotify credentials on startup/when login credentials are missing.

medwards:
Fixes #1 but requires librespot-org/librespot#313 (though I can probably make a workaround)

This PR introduces a login helper interface (authentication.rs) that is invoked if there is no config file or the config file fails to parse. It proposes a breaking change to the config file format to match librespot credentials. *Note:* it also writes to the config file, so back it up if this is a concern.

It requires a bit of work before I'm happy with it:
* improve login interface caller process (it's a mess right now, and doesn't verify the credentials before fully launching ncspot)
  * probably allows the login helper to return `Credentials` instead of its serialization
* Login vs Login /w Facebook have different layouts (to allow URL copy/pasta)
* polling should probably be done with a future instead
* winapi dependency is copy/pasta, probably there are no ncurses backends that support windows so I can remove it

At this point my biggest feedback request is:
* Are we ok changing the config file?
* Is this way of doing logins OK, or would we rather have the login interface part of the main cursive application?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ maintenance = { status = "experimental" }
 [dependencies]
 clap = "2.32.0"
 chrono = "0.4"
+reqwest = "0.9.11"
 crossbeam-channel = "0.3.8"
 directories = "1.0"
 failure = "0.1.3"
@@ -34,12 +35,15 @@ rand = "0.6.5"
 
 [dependencies.librespot]
 git = "https://github.com/librespot-org/librespot.git"
-rev = "2fb901a743906a5b49b3148dbfa85074964dd745"
+rev = "14721f4"
 default-features = false
 
 [dependencies.cursive]
 version = "0.11.1"
 default-features = false
+
+[dependencies.winapi]
+git = "https://github.com/overdrivenpotato/winapi-rs"
 
 [features]
 pulseaudio_backend = ["librespot/pulseaudio-backend"]
--- a/src/authentication.rs
+++ b/src/authentication.rs
@@ -1,0 +1,169 @@
+use std::path::Path;
+
+use cursive::view::Identifiable;
+use cursive::views::*;
+use cursive::{CbSink, Cursive};
+
+use librespot::core::authentication::Credentials as RespotCredentials;
+use librespot::protocol::authentication::AuthenticationType;
+
+pub fn create_credentials(path: &Path) -> Result<RespotCredentials, String> {
+    let mut login_cursive = Cursive::default();
+    let mut info_buf = TextContent::new("Failed to authenticate\n");
+    info_buf.append(format!(
+        "Cannot read config file from {}\n",
+        path.to_str().unwrap()
+    ));
+    let info_view = Dialog::around(TextView::new_with_content(info_buf.clone()))
+        .button("Login", move |s| {
+            let login_view = Dialog::new()
+                .title("Login with Spotify username and password")
+                .content(
+                    ListView::new()
+                        .child("username", EditView::new().with_id("spotify_user"))
+                        .child(
+                            "password",
+                            EditView::new().secret().with_id("spotify_password"),
+                        ),
+                )
+                .button("Login", |s| {
+                    let username = s
+                        .call_on_id("spotify_user", |view: &mut EditView| view.get_content())
+                        .unwrap()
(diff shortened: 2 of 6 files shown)


medwards:
So the config struct got extended while I worked on this. I've rebased so everything is up-to-date at least, but the login helper will wipe out the new settings... (actually maybe every execution, I need to look into it)

Speaking of theming thats the other piece of feedback I'd like on this PR. The login interface's style isn't really in line with the rest of ncspot.

hrkfdn:
Hey there and thanks a lot for your work on this!

> * Are we ok changing the config file?

Yes, I think we are. Especially, because this simplifies the configuration process quite a bit! Maybe it'd make sense to store the login data in a separate file? I was already thinking about creating a folder for configuration files (`~/.config/ncspot/*`) instead of a single file in `~/.config`, which also seems to be more in line with XDG.

> * Is this way of doing logins OK, or would we rather have the login interface part of the main cursive application?

Mhh, if keeping it a separate application simplifies it, I think I'm okay with it. In the end, we don't want to run the main application if we can't login and I don't believe the authentication application will need any of the views/features of the main application, right?

medwards:
> ... I was already thinking about creating a folder for configuration files

I had the same thought, I'll work towards introducing it in this PR
unless you get it done beforehand.

> In the end, we don't want to run the main application if we can't login and I don't believe the authentication application will need any of the views/features of the main application, right?

Agreed on the former, on the latter I need to look into credential
verification. Right now if you enter an invalid username/password the
app panics in an ugly fashion so I'd really like to verify the
credentials and force the user back into the login loop if they're not
valid. If there's an easy librespot call for that, then perfect, but
its possible I'll need to factor out whatever the main app is using.

FYI I got a busy week, but this still has that dependency on my
unmerged librespot PR so no rush.

tyvm for the feedback!

hrkfdn:
@medwards: Small heads up, the config file now resides in `~/.config/ncspot/config.toml`.

medwards:
* separated config.toml and credentials.toml
* introduced some config file helpers (write some defaults if the files don't exist, variable callback logic)

Spotify auth endpoint polling is still done without futures and there is no verification of the credentials so entering a bad username/pass still results in ungraceful failure later.

On an unrelated note: I'm a couple of issues and I'm not sure if they're ncspot or upstream projects (ie `librespot` or `rodio`). Can I spam the issue tracker with them until they're nailed down?

hrkfdn:
> On an unrelated note: I'm a couple of issues and I'm not sure if they're ncspot or upstream projects (ie librespot or rodio). Can I spam the issue tracker with them until they're nailed down?

Yeah, go ahead.

Do you still want to polish this in your branch or should we get this in tree?

medwards:
I'm fine with merging if you have no review comments.

medwards:
I might have some sort of obscure bug with this change. On one machine, everything works fine, on the other I get
```
thread '<unnamed>' panicked at 'Authentication failed with reason: BadCredentials', /home/medwards/.cargo/git/checkouts/librespot-06fda9f186b35c32/daeeeaa/core/src/connection/mod.rs:93:21
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Canceled', src/libcore/result.rs:997:5
```

then everything launches and I can play music but the output is some garbled garbage.

I'm going to try to press gang some people to see if its just something weird on my one machine or a broader issue. I would be v happy if you verified this change locally before merging as well XD

hrkfdn:
I tested this after your initial submission and it worked fine. No rush, though. We should probably investigate this. I might be able to work on the polling and credential checking this week, too. I wanted to finish that first before merging it.

medwards:
Eh, the broken audio looks like just a library behind the scenes getting updated now or a file handles issue. All good now after a reboot.

hrkfdn:
I have merged this with some minor wording changes ;)

We can keep working on this in-tree. Thanks again!

medwards:
LOL I completely forgot it was still called `crappy_poller` XD

tyvm for the wording changes

PR #45, opened 2019-03-24
Include new build requirement

medwards:
It took me a little while to figure out it was `libdbus-1-dev` so noting it in the README


hrkfdn:
Merged, thanks!

PR #47, opened 2019-03-25
Add seeking, shuffle and repeat

KoffeinFlummi:
Adds seeking (both with commands/keys and mouse events in the progress bar), shuffle and repeat. All in one PR mostly because I wanted seeking to test it. Let me know if you want me to split these up into multiple PRs.

Since there are no proper Unicode characters for repeat/shuffle, I added a config setting to use NerdFont characters for shuffle, repeat and playback status:

I also reintroduced the PulseAudio backend since Rodio didn't work on my system (other people with similar setups seem to have issues as well: https://github.com/tomaka/rodio/issues/173). I was working on a build script to set a sensible backend at compile time, but I still have some kinks to work out there.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,7 @@ version = "0.11.1"
 default-features = false
 
 [features]
+pulseaudio_backend = ["librespot/pulseaudio-backend"]
 rodio_backend = ["librespot/rodio-backend"]
 portaudio_backend = ["librespot/portaudio-backend"]
 termion_backend = ["cursive/termion-backend"]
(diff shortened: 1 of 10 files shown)


hrkfdn:
Nice work! It seems that shuffle turns on every time a track is added, though. Or am I missing something?

KoffeinFlummi:
Woops. Should be fixed.

KoffeinFlummi:
Oh, I forgot to mention: Since I changed the `duration` attribute of `Track` to contain milliseconds for seeking, you will have to remove your cached playlists to prevent issues, since those still have the old second values.

hrkfdn:
Yeah, I noticed when I saw that all my tracks suddenly had a length of zero ;)
Merged, thanks for the fix!

KoffeinFlummi:
Thanks. Btw, it might be nice to have some sort of Slack/riot.im/IRC/etc setup. I was planning to look into some library browsing/management features and it would probably be best to coordinate that a bit more.

hrkfdn:
Good idea, I've set up a Gitter and sent you an invite

PR #52, opened 2019-03-28
Refactor command handling

KoffeinFlummi:
I refactored the command handling to use a system similar to Cursive's event handling. Commands are sent to views recursively via a new trait before they are handled globally. This implements the original search bar focus behaviour with rebindable keys and makes command handling for currently selected items a bit easier. The default play/queue behaviour is implemented by `ListView` and can be prevented if necessary (e.g. in `QueueView`). This should scale better for more types (artists, albums) and list views.

I also changed the main focus keybindings (F1, F2, F3) to use a seperate `focus` command to avoid unnecessarily polluting the command namespace, so the "x selected" stuff isn't necessary anymore.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -2,39 +2,45 @@ use std::collections::HashMap;
 use std::sync::Arc;
 
 use cursive::event::{Event, Key};
+use cursive::views::ViewRef;
 use cursive::Cursive;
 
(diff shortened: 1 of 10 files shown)


hrkfdn:
Good stuff! Merged with some minor changes.

PR #54, opened 2019-04-04
Only allow owner read/write to credentials.toml

osm:
This PR sets the file mode bits to 600 for credentials.toml. 

The `from_mode` (https://doc.rust-lang.org/std/os/unix/fs/trait.PermissionsExt.html#tymethod.from_mode) function is only available on Unix systems and since I don't have access to a Windows box I decided to set the target_family to unix.

Thanks for an awesome project!


hrkfdn:
Reasonable change! Thank you, merged!

PR #57, opened 2019-04-09
Add support for Spotify URI searches

osm:
With this PR it's now possible to enter Spotify URIs in the search box.

I had to refactor the `process_playlist` function since it expected a `SimplifiedPlaylist` object to be passed and the method that fetches a specific playlist returns it as a `FullPlaylist`.

The refactoring lead me to store the specific values that was used from the playlist on the internal `Playlist` struct instead of the whole object that was return by the Spotify API.

We are only using `id`, `name` and `snapshot_id` from the playlist object, so I think that this should be an OK solution.

### DIFF
--- a/src/playlists.rs
+++ b/src/playlists.rs
@@ -3,7 +3,7 @@ use std::ops::Deref;
 use std::path::PathBuf;
 use std::sync::{Arc, RwLock, RwLockReadGuard};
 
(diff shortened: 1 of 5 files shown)


hrkfdn:
Thanks! I have merged this with some refactoring changes to zap the duplicate cloning that was starting to pile up :)

PR #59, opened 2019-04-19
Implement Library viewing/modification

KoffeinFlummi:
Uses a `TabView` for displaying saved tracks, albums, artists and playlists. Saved/Followed status is displayed in every `ListView` and items can be saved/unsaved;followed/unfollowed via keybinding (default `s`).

Some API endpoints were missing in rspotify, so I implemented those on a fork for now. I've created a pull request upstream though: https://github.com/samrayleung/rspotify/pull/38

Loading tracks/albums is aborted after the first page if that page matches the cache, so with the exception of the first launch or ones where the library was modified on another device, startup is fairly quick. All 4 types are loaded in seperate threads.

(semi-)unrelated changes:
- Fixed keybinding parsing when using modifiers with chars.
- Changed queue saving to use command system and changed default keybinding to `Ctrl+s`.
- Changed play behaviour when playing single tracks to load entire `ListView` into queue to mirror official client behaviour and enable playing entire library.
- Added .log to .gitignore


hrkfdn:
Good stuffs :+1:! Merrrrggeeddddd!

PR #61, opened 2019-04-22
Implement browsing albums, artists and playlists

KoffeinFlummi:
Adds a view stack to `Layout` and commands for opening the selected item and going to the respective album/artist.


hrkfdn:
Coooool! Merged!

PR #63, opened 2019-04-24
Add ALSA backend and fix typo in README.md

osm:
ncspot consumes ~40 % cpu when I use the rodio backend, when I switch to alsa it uses ~0-1% cpu instead.

The ncspot thread that is responsible for the 40 % cpu usage is the audio backend (playback) thread.

With rodio it consumes 40 % even though a song is not playing, but with alsa it just consumes the 40 % cpu during playback. So I prefer to compile it using the alsa backend so my cpu can rest a little when I'm not listening to anything :-)

This occurs on my Carbon X1 6th gen laptop with a Linux 4.15.0-47-generic kernel.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,7 @@ default-features = false
 git = "https://github.com/overdrivenpotato/winapi-rs"
 
 [features]
+alsa_backend = ["librespot/alsa-backend"]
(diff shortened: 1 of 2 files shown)


hrkfdn:
Ah, I noticed an increase in idle load, too, but haven't gotten around to investigating it yet. Merged, thanks!

PR #64, opened 2019-04-25
Use ALSA backend with the Vorbis decoder

osm:
By switching to the ALSA backend with the Vorbis decoder the CPU load went from ~40 % load during playback to ~14 %

The `rodio` backend combined with the Vorbis decoder got the load down from ~40 % to ~30 %.

I also tested with the `tremor` decoder (https://github.com/librespot-org/librespot/blob/master/Cargo.toml#L66) and the ALSA backend but that produced ~20 % CPU load.

So, at least on my computer the ALSA backend with the Vorbis decoder seems to be the best combination since it just takes up ~14 % CPU load.


hrkfdn:
Hi and thanks for the PR. Moving to ALSA by default seems problematic to me in terms of portability, as Rodio/CPAL handle playback for multiple operating systems instead of just Linux. Also, it's the default backend for librespot now.

This definitely needs to be fixed at some point, though. Is upstream aware of this problem?

osm:
That's a good point. So maybe we can keep Rodio/CPAL as default playback but still use the Vorbis decoder since it reduces the load with ~10 %. I can change my PR to be like that instead if you want to.

I have just briefly checked their issues section but couldn't find anything similar to this problem though, so I'm not sure if they are aware of it.

plietar:
Are you sure you are building in release mode? Lewton’s performance is quite bad in debug, but comparable to libvorbis in release

hrkfdn:
Mhh, I think something is off. The CPAL event loop is consuming the majority of CPU cycles, and this is even without playing a song:

https://affekt.org/tmp/ncspot_prof.svg

hrkfdn:
A release build produces a ~6% CPU usage. PulseAudio goes up to 11% at the same time. All of this without playing a song. I'm guessing the event loop unnecessarily bothers PulseAudio.

j605:
It would be nice to reuse the work done in `mpv` and directly use it as a player. One can also use it as a library (`libmpv`) and maybe get audio handling across multiple platforms for free.

Issue #74, opened 2019-05-17
Move to strongly typed message passing


Herbstein:
Right now the program uses a string-based message-passing system. The primary reasons for this is enabling exhaustive pattern matching, and avoiding having to box and unbox strings internally. 

The current command system (e.g. `:<command> <args>`) slots neatly into the internal messages, which is currently a big plus. But I do think having a small command-translation layer is a small price to pay for otherwise proper internal message passing.

Herbstein:
I've been posting most things that I notice could be improved. If I start becoming annoying or a pest do let me know :)

I've been using this as my daily driver, and have been loving it. i

hrkfdn:
Hey,

could you elaborate on this a little? What would be the benefit? Less error-prone command handling in the UI parts? I agree that would be an improvement. In the beginning, commands were passed as enums via messages, but I believe we migrated to a string based approach to allow the user to directly input commands in the prompt and enable future keybindings.

> I've been posting most things that I notice could be improved. If I start becoming annoying or a pest do let me know :)
> 
> I've been using this as my daily driver, and have been loving it. i

Oh not at all, thanks for all the feedback.

Herbstein:
Rust has a very strong type system. I think it's a shame if we forgo it to simply throw around strings until the last minute. Instead, I propose commands become strongly typed as early as possible. For internally passed commands - like for keybinds - this would mean they are always strongly typed.

For commands you'd need a parser. It's quite a simple parser that can split on spaces. 

I'm sure there is a good reason for going away from the Enum-based system, but I personally would prefer a stronger alternative to the current implementation.

hrkfdn:
That's a reasonable approach. Are you interested in working on this?

hrkfdn:
Fixed via https://github.com/hrkfdn/ncspot/pull/76

PR #76, opened 2019-05-20
WIP: Move to enum-based commands

Herbstein:
Exploration of needed changes for #74.

Biggest needs right now is handling default command behavior and the command line - both are completely disabled at this time.

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -1,4 +1,6 @@
 use queue::RepeatSetting;
+use std::collections::HashMap;
+use std::iter::FromIterator;
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub enum PlaylistCommands {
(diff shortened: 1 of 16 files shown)


Herbstein:
Everything here is subject to change. I have basically pulled out the old commands into enums. 

Some of the commands, like the `Move` and `Shift` commands, have two parameters but a "strong" default. I represented those second parameters with `Option<T>` types, but that quickly becomes unwieldy. Another way I imagined doing this was having the associated `-Mode` enum have both a `Down` and a `Down5` element - but that is even less ergonomic.

Not entirely sure if we should just leave them as they are now, or what the right change is.

Herbstein:
The biggest problem with implementing the default command behavior is that the callbacks are matched over the string representations. Do you have a suggestion as to how we would rewrite that part?

I am personally considering if the `CommandManager` struct should have `Arc`'s to the three Spotify structs, but if we can avoid that it'd be great. Another option is to have a single callback created and stored in the `register_all` call - the callback can then do a match. This shouldn't be too bad of an overhead because we're using enums.

hrkfdn:
Great work, thank you!

I think using `Option<T>` for the `Move` and `Shift` commands is okay. It may be a bit cumbersome, but it's cleaner than hardcoding them and more flexible when we need to parse the commands from binding configurations at some point.

The callback variant is fine. I don't think overhead is a problem here, as the commands are only processed on user interaction, which doesn't happen often enough for it to cause a worrying overhead, especially since we would we using enums, as you said.

On the other hand, if keeping `Arcs` simplifies things greatly, feel free to do it. They're more like global objects anyway.

Herbstein:
I decided to go with the `Arc`s. It becomes a much cleaner implementation and it reduces the amount of references to the structs. 

The last thing missing is a parser for the commandline. It shouldn't be too hard to get something implemented there too so I'll get that done.

Herbstein:
I just looked at it a bit more, and I'm wondering if we even keep the string-based keybinds? The parsing functions are great for parsing external files, but since both the `Event` enum and the `Key` enum are hashable we could simply put those enums into the `HashMap` instead of moving back and forth from strings.

Herbstein:
I did some smoke testing, and it looks like the command parser is working as expected.

Herbstein:
Any entered command that isn't correct is just silently ignored atm. The goal should be to print out "unknown command" as previous, so let's look into that next?

hrkfdn:
That's not a hard requirement :) We can take care of that in-tree, as well.

Herbstein:
Then I think this might be done.

PR #79, opened 2019-05-30
Sketch new Queue struct

### LINKED ISSUE #69: Use a two-queue system like the official clients
The official Spotify clients use, effectively, two queues. The first queue is the general playback queue, which would be what `ncspot` has now. The second queue is the user-handled queue. This is the queue a song is added to when you press "Add to Queue" in the official clients.

I took a screenshot from the Linux client to illustrate.

The user queue is emptied first, and then the playback queue is used. 

It's a feature that is great when you're playing a playlist, but want to hear a specific song next. This could be because you saw it mentioned online, a friend mentioned it, or any other reason.

Herbstein:
This is an effort to explore how to fix #69. There's quite a few challenges here, but I hope to simplify the `Queue` interface significantly along the way.

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -3,6 +3,9 @@ use std::sync::{Arc, RwLock};
 use rand::prelude::*;
 
 use spotify::Spotify;
+use std::collections::VecDeque;
+use std::fmt::{Debug, Display};
+use std::mem;
 use track::Track;
 
 #[derive(Clone, Copy, PartialEq, Debug, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


Herbstein:
The current functionality matches the official clients. Still missing _a lot_ but working within the current framework seems like a good idea IMO

Herbstein:
This is right in the middle of my exams and implementing a new queue system would be too time consuming at the moment. I hope I can get back to this.

The major problem I am facing is that the `Queue` struct is used everywhere, and there are some leaky abstractions that (potentially) could be abstracted away. Leaving the pr here for now, and I'll see when I get some time.

hrkfdn:
No worries, it won't run away. Good luck with your exams! Maybe I can
contribute to this as well when I have more time.

PR #120, opened 2020-01-21
Changed 'seek' keybinding to match ncmpcpp and added fast-seek

### LINKED ISSUE #104: Improve seek command
I stopped using MPD for a while and used MPV instead. They have a really great solution to seek quickly:

From their Wiki:
> **LEFT and RIGHT**
    Seek backward/forward 5 seconds. Shift+arrow does a 1 second exact seek (see --hr-seek).
**UP and DOWN**
    Seek forward/backward 1 minute. Shift+arrow does a 5 second exact seek (see --hr-seek).

I think this is a great concept to have a 'fast seek' and 'slow seek' option. Using the current 'seek' command (jump 1 second ahead) can be quite cumbersome, especially for long tracks.

I you don't like the idea with quick/fast seek, please consider increasing the jump to something like 5 seconds.

I know you can use the mouse to jump to the exact position, but terminal app users try to avoid using the mouse ;)

freijon:
Since you compare your app to 'mpd', I thought it would make sense to align the keybindings for the seek command with ncmpcpp, which is probably the most common mpd client:
- 'f' to seek forward
- 'b' to seek backward

I also implemented a fast-seek option, which would fix #104

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,8 @@ have them configurable.
 * `Shift-s` stops a track
 * `Shift-u` updates the library cache (tracks, artists, albums, playlists)
 * `<` and `>` play the previous or next track
-* `,` and `.` to rewind or skip forward
+* `f` and `b` to seek forward or backward
+* `Shift-f` and `Shift-b` to seek forward or backward in steps of 10s
 * `-` and `+` decrease or increase the volume
 * `r` to toggle repeat mode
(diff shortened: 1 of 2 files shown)


hrkfdn:
Cool, thank you! I personally like this, but ncspot needs configurable keybindings soon, no way we can keep everyone happy :smile:

PR #124, opened 2020-02-03
added docker-compose.yml to build binary for linux

Eun:


hrkfdn:
Is this required for some sort of packaging setup? Or meant as a help for users? I like the idea, but I'm unsure if we can come up with a one size fits all solution :thinking:

Eun:
It’s just an easy way for people to build the application.
I guess we can add more targets in docker-compose in the future

whisperity:
Oh, it seems multiple people came up with the idea of using Docker to self-compile the binary. I made a similar script from scratch, too. There isn't a "binary release" available for the project. It's either package managers for other distros (Debian/Ubuntu people seem not to be supported, unless they want to use the snapd version...), or having to self-compile which involves installing Rust toolkit on your own machine.

Thus, pre-building the binary in Docker and then just throwing it into your bindir, and installing I think two additional libraries you can get it to work nice and fair, no snaps and such needed.

whisperity:
I believe if you do not intend to support Docker, a binary release could be uploaded together with the releases by GitHub Actions in the first place.

hrkfdn:
I think that's a good idea. Superseded by #393.

PR #126, opened 2020-02-12
Add snap build info

popey:
Hello! I added ncspot to the snap store - see https://snapcraft.io/ncspot - which has automated builds for i386, amd64, arm64, ppc64el and s390x (know anyone with an IBM mainframe who wants to listen to Spotify on it? :D )

This merge simply adds the build status as you have for repology. Sorry it doesn't line up. I'll start a conversation with repology to see if they can support snapcraft too. 

Thanks for making a great spotify client <3


hrkfdn:
Cool, thanks for this!

PR #127, opened 2020-02-20
Added configuration option for initial shuffle, repeat, volume

dann-merlin:
I was a bit annoyed, that by default ncspot starts with shuffle and repeat turned off and volume at 100% so I thought why not change that.
This change lets ncspot read these options from the config.toml:
```
[saved_state]
volume = 0 - 65535
repeat = "track" | "playlist" | anything else defaults to no
shuffle = true | false
```

Those values are read at the start of the program and set the initial values of the respective variables.

Anyways, I love the app and I'm happy to contribute!

In the future I might write something to sync changes to these values made in the app with what's in the config file

PS: My first time writing in rust. Hope it's okayish code. There were many confusing hurdles for me as a newbie.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -12,6 +12,14 @@ pub struct Config {
     pub keybindings: Option<HashMap<String, String>>,
     pub theme: Option<ConfigTheme>,
     pub use_nerdfont: Option<bool>,
+    pub saved_state: Option<SavedState>,
+}
+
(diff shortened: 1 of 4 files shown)


hrkfdn:
Nice, thanks! I'm wondering if we should use a range from 0-100 for the configurable volume instead. What do you think?

dann-merlin:
Yeah, you are totally right. Updated it!

hrkfdn:
Merged, thank you!

PR #128, opened 2020-02-20
Shuffle: Start with a random track if not selecting track directly

junglerobba:
Starting a playlist, album or artist with shuffle enabled always
starts on the first track in it and then plays the rest of the
queue shuffled.
This changes it so unless a track is picked directly, playback
will start on a random track of the selection.

This matches the behavior of most music players that I am aware of,
as well as the official Spotify client.


hrkfdn:
Good work, thanks!

PR #142, opened 2020-02-29
removed unnecessary parentheses

AlwaysLivid:
This one's a small change that doesn't really affect functionality and just gets rid of a warning during building, but ¯\_(ツ)_/¯

Buillding the program was successful on my end.


hrkfdn:
Merged, thanks!

PR #147, opened 2020-03-03
added a script that compiles the project inside docker debian container

emdo-pa:


Zocker1999NET:
I think this PR tries to accomplish the same goal as #124

hrkfdn:
Superseded by #393

PR #156, opened 2020-03-18
Actually set volume when reading in from config

dann-merlin:
Hey, this is actually just a fix to a "bug" that just spawned from me not properly testing my stuff. The volume set in the config file was not set properly on the initial start of the program. So this is literally just a one-liner.


hrkfdn:
Merged, thanks!

PR #167, opened 2020-03-30
Add support for 'home' and 'end' keys

tinez:


BachoSeven:
Nice! I had addressed this in my config by binding "g" and "G" to move up/down 800, but this is better. Do you think I could change the keybindings for this somehow? I mean being able to remap "home" and "end" as you defined them from the user config.

tinez:
> Nice! I had addressed this in my config by binding "g" and "G" to move up/down 800, but this is better. Do you think I could change the keybindings for this somehow? I mean being able to remap "home" and "end" as you defined them from the user config.

I wasn't totally happy with the first version of the patch, but in the current one you can map commands like "move top", "move bottom" to any keys.

hrkfdn:
Thanks for the reiteration, I like it!

PR #174, opened 2020-04-07
README: Added doc for initial / save state

Zocker1999NET:
Documents the feature introduced by #127 for configuring the initial state (for volume, repeat and shuffle) in the README.


hrkfdn:
Appreciated, thanks!

PR #185, opened 2020-04-27
Added OpenUri D-BUS MPRIS support.

Bettehem:
I have been working on a music bot project, and recently I found the ncspot project and wanted to add support to play Spotify songs via ncspot, because the regular Spotify client eats up so much RAM, the bot doesn't always work properly on low-end machines.
All was good until I finished implementing the new functionality and then I noticed that even though ncspot supports most of the mpris protocol, OpenUri method call was missing, which is one of the most important ones for my bot.
I then took a crash course in Rust and implemented it myself. You might want to review my code since I have never done Rust until now...


hrkfdn:
Cool, thanks, merged!

PR #207, opened 2020-06-25
[feat] add Command::Noop

hhsnopek:
## Details
This pull request introduces a Noop command allowing the user to override default keybindings.

### Example
In the configuration a user can override a keybinding in the following way:
```toml
[keybindings]
"s" = "noop" # Disable default save keybinding
"d" = "noop" # Disable default delete keybinding
```

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -102,11 +102,13 @@ pub enum Command {
     Shift(ShiftMode, Option<i32>),
     Search(String),
     Help,
+    Noop,
 }
 
 impl fmt::Display for Command {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let repr = match self {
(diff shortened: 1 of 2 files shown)


src/command.rs
```diff
@@ -102,11 +102,13 @@ pub enum Command {
     Shift(ShiftMode, Option<i32>),
     Search(String),
     Help,
+    Noop,
 }
 
 impl fmt::Display for Command {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let repr = match self {
+            Command::Noop => "no operation".to_string(),
```
hrkfdn, line 111: This representation should resolve to the command name instead

hhsnopek: resolved and rebased into d33147c

PR #211, opened 2020-06-28
enh: add methods and api usage to delete track from playlist

### LINKED ISSUE #209: question: how to remove a track from playlist
i'm kinda shocked that i couldn't find anything by searching for "delete/remove". (am i really the only one stumbling across this?)

how can i remove a track from a playlist?

the context menu (`shift+o`) allows only to add a track to a playlist. but when being in a playlist i also want to be able to remove a song from the playlist.

thomasrstorey:
An attempt to fix #209 
Keep in mind I am extremely new to rust so please do not hesitate to correct me where I am making heinous mistakes! :)
So far I have it making the correct API request when you hit `d` on a track in a playlist, but it does not update the playlist view until you backspace out of the playlist and `o` back into it. Not sure what the disconnect there is.

### DIFF
--- a/src/library.rs
+++ b/src/library.rs
@@ -512,6 +512,27 @@ impl Library {
         }
     }
 
+    pub fn playlist_remove_tracks(&self, playlist_id: &str, track_pos_pairs: &[(Track, usize)]) {
+        let mut has_modified = false;
+
+        if self.spotify.delete_tracks(playlist_id, track_pos_pairs) {
+            let mut playlists = self.playlists.write().expect("can't writelock playlists");
+            if let Some(playlist) = playlists.iter_mut().find(|p| p.id == playlist_id) {
+                if let Some(tracks) = &mut playlist.tracks {
+                    for (_track, pos) in track_pos_pairs {
+                        tracks.remove(*pos);
+                        has_modified = true;
+                    }
+                }
+            }
+        }
+
+        if has_modified {
+            // TODO: Apparently this does not actually update the list of tracks in the ui...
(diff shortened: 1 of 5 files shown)


hrkfdn:
Looks pretty good so far. On a first glance I'd suggest moving the deletion code from library.rs to the `Playlist` model and hand in the `Spotify` reference instead. Within the `Playlist` object, if you delete the entries from the `tracks` Vec the UI should update automatically.

thomasrstorey:
Thanks! I actually did have the code in the Playlist model when I first started, but then I saw that the most analogous function `playlist_append_tracks` lived in `Library` so I moved it over there. I'll try again from inside `Playlist` and see if it's a more natural fit.

thomasrstorey:
Moved the delete handling code into `Playlist` and it did get a lot simpler as a result, but did not immediately solve the problem of the ui not updating.
Turns out I needed to mutate the `ListView.content` directly, I think because the `content` is getting initialized from a new `Arc<RwLock<Vec<Track>>>` in the `new`of `PlaylistView`, as compared to the `QueueView` which passes a `clone()` of the `queue` in? I'm not totally sure, please correct me if mutating the `ListView.content` directly is Very Bad for Reasons.

hrkfdn:
Merged with some very minor changes, thank you! :)

thomasrstorey:
Awesome! Thanks for giving me the space to contribute to a such a fun project. And for having such well-organized code that made it easy for me to figure out where my changes should go!

hrkfdn:
My pleasure! Thanks for your contribution!

PR #214, opened 2020-07-02
Add comment to use `--locked`

daniel-vera-g:
As described in: https://github.com/hrkfdn/ncspot/issues/212#issuecomment-652055797

PR #215, opened 2020-07-04
Fix "workarround" typo for "workaround"

daniel-vera-g:
See https://github.com/hrkfdn/ncspot/pull/214#issuecomment-653776631


hrkfdn:
Thanks!

PR #222, opened 2020-07-14
Fix mpris metadata returing Spotify URI instead of an url in xesam:url field.

Bettehem:
Hello, I noticed that when getting metadata on the currently playing track, `xesam:url` returns the Spotify URI instead of an url.
I changed this so now it returns an url, which matches with the official client's functionality.


Bettehem:
I have to add, that I'm very new to Rust, so I don't know if this is the best way to solve the issue, but let me know!

src/mpris.rs
```diff
@@ -118,7 +118,16 @@ fn get_metadata(playable: Option<Playable>) -> Metadata {
     );
     hm.insert(
         "xesam:url".to_string(),
-        Variant(Box::new(playable.map(|t| t.uri()).unwrap_or_default())),
+        Variant(Box::new(
+            playable
+                .map(|t| {
+                    format!(
+                        "https://open.spotify.com/track/{}",
```
hrkfdn, line 125: This will not work for podcast episodes, as the path changes. You could grab the `share_url()` which should give you the correct URL for tracks/podcasts.

Bettehem: Okay, I will look in to this. Should `mpris:trackid` also be changed since it also has `spotify:track:` statically typed or does it stay the same with podcasts?

hrkfdn: Yes, good catch!

Bettehem:
Should be fixed now!

src/mpris.rs
```diff
@@ -118,7 +115,13 @@ fn get_metadata(playable: Option<Playable>) -> Metadata {
     );
     hm.insert(
         "xesam:url".to_string(),
-        Variant(Box::new(playable.map(|t| t.uri()).unwrap_or_default())),
+        Variant(Box::new(
+            playable
+                .map(|t| {
+                    t.share_url().unwrap_or_else(|| "0".to_string())
```
hrkfdn, line 121: This will set `xesam:url` to `0` if no URL is set. Is this according to specs?

Bettehem: I'm not sure, how would I check that?
I haven't ever seen Spotify not returning an url

Bettehem: Perhaps it would be better to just set `xesam:url` to an empty string if no url is found?

Bettehem: I just checked, and when getting metadata for podcasts, `mpris:artUrl` and `xesam:artist` is just empty, no "0" or anything, so `xesam:url` should probably do the same

hrkfdn: Empty string sounds more appropriate than `"0"`.

Bettehem: Yeah, I'll change it.

Bettehem:
Okay, should be fixed now.

hrkfdn:
Merged, thank you!

PR #252, opened 2020-08-20
Change track's position using MPRIS

r4v3n6101:
Implementation of Seek & SetPosition methods of MPRIS interface described here: https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html


hrkfdn:
Nice work, thanks!

Issue #221, opened 2020-07-14
Failing to build on NetBSD


Ramiferous:
Hopefully someone running `ncspot` on OpenBSD can help me here. I'm really keen to get this working on my NetBSD machine but I get a build error:

```
$ cargo run --no-default-features --features portaudio_backend,cursive/pancurses-backend 
...
error: failed to run custom build command for `ncurses v5.99.0`

Caused by:
  process didn't exit successfully: `/home/dave/Downloads/ncspot/target/debug/build/ncurses-cca56809b8b206f2/build-script-build` (exit code: 101)
--- stdout
cargo:rerun-if-env-changed=PKG_CONFIG_PATH
cargo:rustc-link-lib=ncursesw
OPT_LEVEL = Some("0")
TARGET = Some("x86_64-unknown-netbsd")
HOST = Some("x86_64-unknown-netbsd")
CC_x86_64-unknown-netbsd = None
CC_x86_64_unknown_netbsd = None
HOST_CC = None
CC = None
CFLAGS_x86_64-unknown-netbsd = None
CFLAGS_x86_64_unknown_netbsd = None
HOST_CFLAGS = None
CFLAGS = None
CRATE_CC_NO_DEFAULTS = None
DEBUG = Some("true")
CARGO_CFG_TARGET_FEATURE = Some("fxsr,sse,sse2")

--- stderr
/home/dave/Downloads/ncspot/target/debug/build/ncurses-c07513f36080cdf7/out/chtype_size.c:6:10: fatal error: ncurses.h: No such file or directory
 #include <ncurses.h>
          ^~~~~~~~~~~
compilation terminated.
thread 'main' panicked at 'assertion failed: command.status().expect("compilation failed").success()', /home/dave/.cargo/registry/src/github.com-1ecc6299db9ec823/ncurses-5.99.0/build.rs:105:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

warning: build failed, waiting for other jobs to finish...
error: build failed
```

Poraudio & ncurses are installed

```
$ pkgin list | grep portaudio                                                            
portaudio-190600.20161030nb5 Portable cross-platform audio API
$ pkgin list | grep ncurses                                                              
ncurses-6.2nb2       CRT screen handling and optimization package
```

sloshyyy:
[...]

PR #256, opened 2020-08-24
Use libc for setlocale() to fix non-ASCII

voidpin:
character printing.
https://github.com/hrkfdn/ncspot/issues/221

Patch by rvp and kamil to fix character printing.
Fixes issue #221

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,7 +95,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "alsa-sys 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "bitflags 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "nix 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -104,7 +104,7 @@ name = "alsa-sys"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "pkg-config 0.3.18 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -150,7 +150,7 @@ version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "hermit-abi 0.1.15 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -171,7 +171,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "addr2line 0.13.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "cfg-if 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "miniz_oxide 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "object 0.20.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "rustc-demangle 0.1.16 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -354,7 +354,7 @@ version = "0.29.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "glob 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "libloading 0.5.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -455,7 +455,7 @@ version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "core-foundation-sys 0.7.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -463,7 +463,7 @@ name = "core-foundation-sys"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -497,7 +497,7 @@ dependencies = [
  "core-foundation-sys 0.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "coreaudio-rs 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "stdweb 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
@@ -592,7 +592,7 @@ dependencies = [
  "cursive_core 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "enumset 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "maplit 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "pancurses 0.16.1 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -614,7 +614,7 @@ dependencies = [
  "enum-map 0.6.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "enumset 1.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "num 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "owning_ref 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -693,7 +693,7 @@ name = "dbus"
 version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "libdbus-sys 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -742,7 +742,7 @@ name = "dirs"
 version = "1.0.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "redox_users 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
@@ -752,7 +752,7 @@ name = "dirs-sys"
 version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "redox_users 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
@@ -885,7 +885,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "cfg-if 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "crc32fast 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "miniz_oxide 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -1041,7 +1041,7 @@ version = "0.1.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "cfg-if 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "wasi 0.9.0+wasi-snapshot-preview1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -1111,7 +1111,7 @@ name = "hermit-abi"
 version = "0.1.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -1329,7 +1329,7 @@ name = "iovec"
 version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -1394,7 +1394,7 @@ dependencies = [
 
 [[package]]
 name = "libc"
-version = "0.2.74"
+version = "0.2.76"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
@@ -1424,7 +1424,7 @@ name = "libpulse-sys"
 version = "0.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -1505,7 +1505,7 @@ dependencies = [
  "byteorder 1.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "cpal 0.8.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures 0.1.29 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "libpulse-sys 0.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "librespot-audio 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "librespot-core 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -1572,7 +1572,7 @@ name = "malloc_buf"
 version = "0.0.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -1643,7 +1643,7 @@ dependencies = [
  "fuchsia-zircon-sys 0.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "iovec 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "kernel32-sys 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "miow 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "net2 0.2.34 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -1657,7 +1657,7 @@ version = "0.6.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "iovec 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "mio 0.6.22 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -1694,7 +1694,7 @@ version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "openssl 0.10.30 (registry+https://github.com/rust-lang/crates.io-index)",
  "openssl-probe 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -1721,6 +1721,7 @@ dependencies = [
  "futures 0.1.29 (registry+https://github.com/rust-lang/crates.io-index)",
  "futures 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "librespot-core 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "librespot-playback 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "librespot-protocol 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -1747,7 +1748,7 @@ version = "5.99.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "cc 1.0.58 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "pkg-config 0.3.18 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -1757,7 +1758,7 @@ version = "0.2.34"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "cfg-if 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -1768,7 +1769,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "bitflags 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "cfg-if 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.2.74 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.76 (registry+https://github.com/rust-lang/crates.io-index)",
  "void 1.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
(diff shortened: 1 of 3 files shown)


hrkfdn:
Hey, thanks for taking a look at this.

However, I'm not sure if this should be fixed on application level, as it doesn't seem to affect platforms other than NetBSD and will override `LC_ALL` for all users. Could this be done within pkgsrc, maybe?

krytarowski:
It's a generic bug. `""` does not override the native environment.

krytarowski:
Alternatively you can try to patch cursive to use proper wchar functions.

krytarowski:
pkgsrc is also portable to all rust aware platforms, and patching in pkgsrc is not a solution for brokenness in one environment.

hrkfdn:
I appreciate the patch, but I don't believe this is the right approach.

If this is a bug in cursive, it should be fixed there. Have you guys tried the `ncurses` backend instead of `pancurses`?

krytarowski:
We use ncurses, but cursive uses non-wide APIs for wide input.

hrkfdn:
What I meant is that cursive has multiple backends to write to terminals. By default, ncspot uses pancurses. But there is also an ncurses backend, which does not use the pancurses library. It also seems to set the locale as in this patch:

https://github.com/gyscos/cursive/blob/ad935f5126ed5ac763b57e3c08a73bd3d69676df/cursive/src/backends/curses/n.rs#L90

Maybe it's worth a try to switch from `cursive/pancurses-backend` to `cursive/ncurses-backend`.

krytarowski:
Please show a patch and we can test it.

hrkfdn:
You're already patching Cargo.toml, just switch the backend like so:

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 2c55e9b..9abad74 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,4 +56,4 @@ rodio_backend = ["librespot-playback/rodio-backend"]
 portaudio_backend = ["librespot-playback/portaudio-backend"]
 termion_backend = ["cursive/termion-backend"]
 mpris = ["dbus"]
-default = ["share_clipboard", "pulseaudio_backend", "mpris", "cursive/pancurses-backend"]
+default = ["portaudio_backend", "cursive/ncurses-backend"]
```

See also this comment: https://github.com/hrkfdn/ncspot/issues/221#issuecomment-675903920

rajeevvp:
Hi, I'm the `rvp` mentioned in the first post at the top.

> I appreciate the patch, but I don't believe this is the right approach.

No, this is the correct approach. As per the POSIX standards, all programs start by default in the `C` aka `POSIX` locale. In this locale, all input and output are assumed to consist of single byte chars, and the _printable_ ones in these are the standard ASCII ones.

If you input or output UTF-8 multi-byte chars., as you are doing, then you _must_ call `setlocale(3)` to inform `curses(3)` of this fact, otherwise you get that characteristic mojibake in the screenshot. And `setlocale(LC_ALL, "");` is the _standard_ way of setting the user's specified locale.

The way this works is:
1. The user sets his preferred locale, typically, using environment variables. eg.
```
export LANG=en_GB.UTF-8
export LC_CTYPE=en.GB.UTF-8
...
```
2. The program starts up in the `C` locale by default.

3. `setlocale(LC_ALL, "")` is called by the program--_which sets the user's preferred locale_. Otherwise, the program _remains in the `C` locale_.

4. The programmer can override the user's locale settings by calling `setlocale(TYPE, NON_NULL_VALUE)` explicity.


In fact, the example programs supplied with `pancurses` and `ncurses-rs` explicitly call the Rust version of `setlocale`. More than that, the libraries themselves, both of 'em, call the Rust equivalent of `setlocale(LC_ALL, "")` on initialization. What's puzzling to me--not being a Rust programmer, I can't follow the code very well--is _why it's not taking effect_ on NetBSD.

As I mentioned, I'me not a Rust guy, but, in investigating this little issue for the UnitedBSD forum, I dusted up an old C curses test program I had lying around and updated it for UTF-8. It's a tad too long for test-case code, but, it's what I used.

Run the binaries with the `-b utf` flag and with `-n`. This skips calling `setlocale(LC_ALL, "")` and outputs raw UTF-8 resulting in the same garbled box-printing characters as in Rust. Interestingly, NetBSD's `curses(3)` library calls `setlocale` internally, so `-n` has no effect. `ncurses` leaves that up to the application program, which is why you get wonky chars in both Rust and my programs which link with `ncurses`. Finally, statically compiled programs, produce mojibake on both Linux and NetBSD because `setlocale()` doesn't work in static binaries. This is mentioned in the `setlocale(3)` man-page.

Makefile:
```Makefile
NINC = -DDO_NCURSES -I/opt/ncurses/include
NLIB = -L/opt/ncurses/lib -lncursesw -ltinfow
CINC =
CLIB = -lcurses

SRC = msgbox.c
EXE = ncurses_shared ncurses_static netbsd_curses_shared netbsd_curses_static

all: ${EXE}

ncurses_shared: $(SRC)
	${CC} ${NINC} ${SRC} -o $@ ${NLIB}

ncurses_static: ${SRC}
	${CC} -static ${NINC} ${SRC} -o $@ ${NLIB}

netbsd_curses_shared: ${SRC}
	${CC} ${CINC} ${SRC} -o $@ ${CLIB}

netbsd_curses_static: ${SRC}
	${CC} -static ${CINC} ${SRC} -o $@ ${CLIB} -lterminfo

clean:
	rm -f ${EXE}
```

Test program:
```c
/**
 * Show text inside a box in the middle of the screen.
 */
#include <errno.h>
#include <locale.h>
#include <setjmp.h>
#include <signal.h>
#include <stdarg.h>
#include <stdlib.h>
#include <string.h>
#include <unistd.h>			/* for alarm() */
#ifdef DO_NCURSES
#include <ncursesw/ncurses.h>
#else
#include <curses.h>
#endif

/* Functions */
static char* addstring(char* t, char* s);
static char* mksp(int* colno);
static int do_opts(int argc, char* argv[]);
static int maxlen(char** v, int nl);
static int mklines(char*** v, char* msg);
static void do_setup(void);
static void error(char* fmt, ...);
static void freev(char** v, int nl);
static void msgbox(char* msg, int wait);
static void mybox(WINDOW* w);
static void sig_hand(int signo);
static void* Realloc(void* ptr, size_t size);

enum box_type {
	BT_ASC = 1,
	BT_DEF,
	BT_UTF
};

/* Globals */
static char* prog;			/* program name */
static struct options {			/* program options */
	enum box_type bt;		/* type of char. used to draw box borders */
	int do_locale;			/* call setlocale(3) if 1 */
	int wait;			/* show text for wait seconds */
} opts;
static char* msg;			/* pointer to message text */
static sigjmp_buf jmpbuf;		/* Env. for sigsetjmp/siglongjmp */
static volatile sig_atomic_t canjump;	/* jmpbuf is valid if 1 */

/**
 * M A I N
 */
int
main(int argc, char* argv[])
{
	int argidx;

	prog = argv[0];

	argidx = do_opts(argc, argv);
	if (argidx >= 0) {
		argc -= argidx;
		argv += argidx;
		if (argc == 1 && **argv)	/* have user text */
			msg = *argv;
		else			/* help text */
			msg = "Usage: msgbox [ -b type ] [-n] [-t secs] text\n"
			"Show ASCII text in a message box like this one.\n"
			"\n"
			"  -b type    Box drawing char type: ascii, default, utf.\n"
			"  -n         Don't call setlocale(3)--for testing.\n"
			"  -t secs    Show text for `secs' before exiting.\n"
			"\n"
			"The text is, by default, displayed for 5 seconds.\n"
			"\n"
			"Press (almost) any key to exit early.";
	}
	
	/* w/o this ncurses displays mojibake on raw UTF-8 output */
	if (opts.do_locale)
		setlocale(LC_ALL, "");
	do_setup();
	msgbox(msg, opts.wait);
	/* Shut down curses */
	endwin();
	return 0;
}

/**
 * Initialize signal handler and curses.
 */
static void
do_setup(void)
{
	/* `jmpbuf' is not setup */
	canjump = 0;

	/* Exit cleanly on signals */
	signal(SIGINT, sig_hand);
	signal(SIGALRM, sig_hand);

	/* Init. curses */
	initscr();
	nonl();
	noecho();
	cbreak();
	keypad(stdscr, TRUE);
	refresh();
}

/**
 * Show the string in `msg' inside a box in the middle of
 * the screen for `wait' seconds.
 * Exit early if any key is pressed.
 */
static void
msgbox(char* msg, int wait)
{
	/* Constants for centering text inside message box */
	const int PAD_X = 4;
	const int PAD_Y = 2;
	const int BEG_X = PAD_X / 2;
	const int BEG_Y = PAD_Y / 2;

	WINDOW* win;
	int maxx, maxy;
	int winx, winy;
	int i, nl;
	char** v;

	/* Sanity check */
	if (msg == NULL || *msg == '\0')
		return;

	/* Get terminal screen size */
	getmaxyx(stdscr, maxy, maxx);

	nl = mklines(&v, msg);
	/* Calculate msgbox size */
	winx = maxlen(v, nl) + PAD_X;
	winy = nl + PAD_Y;

	/* Check if text size is within screen size */
	if (winx > maxx || winy > maxy) {
		/* Free old lines */
		freev(v, nl);
		nl = mklines(&v, "msgbox: sorry, text message exceeds screen size--exiting.");
		/* Calculate new msgbox size */
		winx = maxlen(v, nl) + PAD_X;
		winy = nl + PAD_Y;
	}

	/* Create a message window in the middle of the screen */
	win = newwin(winy, winx, maxy/2 - winy/2, maxx/2 - winx/2);

	/* Draw a visible border around it */
	if (opts.bt == BT_DEF)
		box(win, 0, 0);
	else
		mybox(win);

	/* Display message text inside box */
	for (i = 0; i < nl; i++) {
		wmove(win, BEG_Y + i, BEG_X);
		waddstr(win, v[i]);
	}
	wrefresh(win);

	/* Set alarm and wait for key press */
	if (sigsetjmp(jmpbuf, 1) == 0) {
		canjump = 1;		/* ok to do a siglongjmp now */
		alarm(wait);
		getch();
		alarm(0);		/* turn off on key press */

		/* Erase msgbox if key pressed */
		werase(win);
		wrefresh(win);
	}
	delwin(win);			/* farewall, my sweet */

	/* Free lines */
	freev(v, nl);
}

/**
 * Draw box around window.
 */
static void
mybox(WINDOW* w)
{
	char *H, *V, *TL, *TR, *BL, *BR;

	int i, x, y;

	if (opts.bt == BT_UTF) {
		H = "─";
		V = "│";
		TL = "┌";
		TR = "┐";
		BL = "└";
		BR = "┘";
	} else {		/* ASCII */
		H = "-";
		V = "|";
		TL = "+";
		TR = "+";
		BL = "+";
		BR = "+";
	}
	getmaxyx(w, y, x);
	for (i = 0; i < x; i++) {
		mvwaddstr(w, 0, i, H);
		mvwaddstr(w, y-1, i, H);
	}
	for (i = 0; i < y; i++) {
		mvwaddstr(w, i, 0, V);
		mvwaddstr(w, i, x-1, V);
	}
	mvwaddstr(w, 0, 0, TL);
	mvwaddstr(w, 0, x-1, TR);
	mvwaddstr(w, y-1, 0, BL);
	mvwaddstr(w, y-1, x-1, BR);
}

/**
 * Convert a string in `msg' into an array of
 * individual lines. Return the no. of lines.
 */
static int
mklines(char*** av, char* msg)
{
	int c, col, nl;
	char buf[2];
	char** v;

	c = col = nl = 0;
	v = NULL;

	for (; *msg; msg++) {
		if (nl == c) {
			/* grow vector */
			v = Realloc(v, (c += 1) * (sizeof(char* )));
			v[nl] = NULL;
			v[nl] = addstring(v[nl], "");	/* add empty string */
		}
		switch (*msg) {
		case '\n':
			col = 0;
			nl++;
			break;
		case '\t':
			v[nl] = addstring(v[nl], mksp(&col));
			break;
		default:
			col++;
			buf[0] = *msg;
			buf[1] = '\0';
			v[nl] = addstring(v[nl], buf);
			break;
		}
	}
	*av = v;
	if (nl < c)
		return nl + 1;		/* last line w/o '\n' */
	else
		return nl;
}

/**
 * Return the length of the longest string
 * in a vector.
 */
static int
maxlen(char** v, int nl)
{
	int i, len, max;

	max = 0;
	for (i = 0; i < nl; i++) {
		len = strlen(v[i]);
		if (len > max)
			max = len;
	}
	return max;
}

/**
 * Signal handler jumps back into msgbox for cleanup.
 */
static void
sig_hand(int signo)
{
	/* Jump only if sigsetjump() was done */
	if (canjump == 1) {
		canjump = 0;		/* jmpbuf is invalid */
		siglongjmp(jmpbuf, 1);
	}
}

/**
 * Add string `s' to a string `t' allocating space
 * as necessary.
 */
static char*
addstring(char* t, char* s)
{
	int slen;

	slen = strlen(s) + 1;
	if (t == NULL) {
		t = Realloc(t, slen);
		strcpy(t, s);
	} else {
		t = Realloc(t, strlen(t) + slen);
		strcat(t, s);
	}
	return t;
}

/**
 * Convert a tab to an appropriate no. of spaces.
 */
static char*
mksp(int* colno)
{
	static char spaces[9];

	spaces[0] = '\0';
	do
		strcat(spaces, " ");
	while (((*colno)++ & 7) != 7);
	return spaces;
}

/**
 * Free the elements of a vector and the vector itself.
 */
static void
freev(char* v[], int nl)
{
	int i;

	for (i = 0; i < nl; i++) {
		free(v[i]);
		v[i] = NULL;
	}
	free(v);
}

/**
 * Process program options.
 */
int
do_opts(int argc, char* argv[])
{
	int opt, rc;

	rc = 0;				/* OK */

	/* defaults */
	opts.wait = 5;			/* exit after 5 seconds */
	opts.bt = BT_DEF;		/* draw box borders using the default ACS chars */
	opts.do_locale = 1;		/* call setlocale(3) by default */

	while ((opt = getopt(argc, argv, ":b:t:n")) != -1) {
		switch (opt) {
		case 'b':
			if (strcmp(optarg, "ascii") == 0)
				opts.bt = BT_ASC;
			else if (strcmp(optarg, "default") == 0)
				opts.bt = BT_DEF;
			else if (strcmp(optarg, "utf") == 0)
				opts.bt = BT_UTF;
			else {
				msg = "msgbox: unknown box char. type.\n"
				      "use: ascii, default, or utf.";
				rc = -1;	/* error */
			}
			break;
		case 'n':
			opts.do_locale = 0;
			break;
		case 't':
			opts.wait = atoi(optarg);
			break;
		case ':':
			msg = "msgbox: argument to option missing.";
			rc = -1;	/* error */
			break;
		default:
			msg = "msgbox: ignoring unknown option.";
			rc = -1;	/* error */
			break;
		}
	}

	if (opts.wait <= 0) {
		msg = "msgbox: delay secs must be a positive integer.";
		opts.wait = 5;
		rc = -1;		/* error */
	}

	return (rc < 0 ? rc : optind);
}

/**
 * realloc wrapper.
 */
static void*
Realloc(void* ptr, size_t size)
{
	void* p;

	p = realloc(ptr, size);
	if (p == NULL) {
		error("%s: realloc() failed.", prog);
		exit(EXIT_FAILURE);
	}

	return p;
}

/**
 * Print an error message.
 */
static void
error(char* fmt, ...)
{
	va_list ap;

	fflush(stdout);

	va_start(ap, fmt);
	vfprintf(stderr, fmt, ap);
	va_end(ap);

	if (errno != 0) {
		fprintf(stderr, " %s", strerror(errno));
		errno = 0;
	}
	fprintf(stderr, "\n");
	fflush(stderr);
}
```

hrkfdn:
> No, this is the correct approach.

What I meant is that this should be fixed in the upstream terminal UI library, not in every single downstream application using it, as ncspot is ncurses-agnostic. The abstraction level I'm on is using UI widgets and dialogs, not raw ncurses calls.

I also want to get this fixed, but I don't think this is where it should happen.

rajeevvp:
Maybe my explanation was too lengthy (certainly, that C-code was) and the central point get didn't get across very clearly as a result. To make this clearer:

1. This is not an OS or ncurses or library issue at all. This is simply what _every_ UNIX application, which doesn't input or output straight ASCII, has to do according to the standards. Otherwise, the program will simply not be conforming to the locale environment that the user has set. This is what `setlocale(LC_ALL, "")` does: make the C-library, ncurses, ..., even rust itself, aware of what the user's environment (what chars. are considered alphabets, numbers, or printable; the format for dates and digits; the separators for digits; etc) is.

For example, if I'm in the US, but, want a UK format for dates, I can do:
```
export LANG="en_US.UTF-8"
export LC_CTYPE="en_US.UTF-8"
...
export LC_TIME="en_**GB**.UTF-8"
```
and _every program_ which has `setlocale(LC_ALL, "");` will work correctly. If I want Japanese char. collation and want time to as in Germany:
```
...
export LC_CTYPe="ja_JP.UTF-8"
export LC_TIME="de_DE.UTF-8"
...
```
Here too, `setlocale(LC_ALL, "");` is what's used to make the program conform to what the user wants.

2. So, _all_ programs not assuming ASCII-only I/O _have_ to call `setlocale(LC_ALL, "");` if they want proper program behaviour. Now, where you and I disagree, is _where_ this call is to be made. You feel that it should be done by the underlying library, and certainly, that can be done, but, the `setlocale` call _in that form_ is typically done in the application. Consider the case of the application setting explicit values for some of the categories, then the underlying library then pulling the rug out from underneath by reverting to the environmental default using `setlocale(LC_ALL, "");`. Now the application has one idea for the settings and the library quite another.

Clearer now?

hrkfdn:
Yes, I understand. But the second point is what I was going for: I'm not responsible for printing the wide characters, the library is, which is why I believe it should be fixed there. I would agree if this bug was about processing regional data as you have outlined, but this is not the case.

hrkfdn:
To get back to your previous point: It *should* work. The pancurses library does call `setlocale()`[1], the `ncurses-backend` in cursive calls `setlocale()`[2]. These calls shouldn't even by necessary in ncspot. I'm still very interested in whether switching to `ncurses-backend` fixes the problem on NetBSD. Any luck with this?

[1]: https://github.com/ihalila/pancurses/blob/ee464d79daab6010d79b724ad8d22a3329fb7761/src/unix/mod.rs#L15
[2]: https://github.com/gyscos/cursive/blob/792c4f1638af60aa349db8f006fa4d52b79fa350/cursive/src/backends/curses/n.rs#L90

gyscos:
Possibly of interest: [libc defines `LC_ALL = 6`](https://docs.rs/libc/0.2.76/src/libc/unix/linux_like/mod.rs.html#555), while [ncurses-rs defines `LC_ALL=0`](https://docs.rs/ncurses/5.99.0/src/ncurses/constants.rs.html#281). This is going to cause differences, not sure which one is right about that.

hrkfdn:
Nice find. Looks like it depends on the platform:
```
earth ~/git/libc/src % ag "LC_ALL:"
fuchsia/mod.rs
1509:pub const LC_ALL: ::c_int = 6;

unix/bsd/mod.rs
239:pub const LC_ALL: ::c_int = 0;

unix/haiku/mod.rs
671:pub const LC_ALL: ::c_int = 0;

unix/linux_like/mod.rs
555:pub const LC_ALL: ::c_int = 6;

unix/solarish/mod.rs
826:pub const LC_ALL: ::c_int = 6;

unix/uclibc/mod.rs
589:pub const LC_ALL: ::c_int = 6;

windows/mod.rs
138:pub const LC_ALL: ::c_int = 0;
```

But if I read this right `0` is used for the BSDs, which then again is in line with the definition of `ncurses-rs`. :unamused:

rajeevvp:
> Yes, I understand. But the second point is what I was going for: I'm not responsible for printing the UTF-8 characters, the library is, which is why I believe it should be fixed there.

But, you are passing raw UTF-8 yourself, too:
```
$ fgrep PlayerEvent ncspot-master/src/ui/statusbar.rs
use crate::spotify::{PlayerEvent, Spotify};
                PlayerEvent::Playing => "\u{f909} ",
                PlayerEvent::Paused => "\u{f8e3} ",
                PlayerEvent::Stopped | PlayerEvent::FinishedTrack => "\u{f9da} ",
                PlayerEvent::Playing => "▶ ",
                PlayerEvent::Paused => "▮▮",
                PlayerEvent::Stopped | PlayerEvent::FinishedTrack => "◼ ",
$ fgrep printer.print ncspot-master/src/ui/statusbar.rs
        printer.print(
            printer.print(
            printer.print((1, 1), &state_icon);
            printer.print((0, 0), &"┉".repeat(printer.size.x));
                printer.print((4, 1), &t.to_string());
            printer.print((offset, 1), &right);
                printer.print((0, 0), &"━".repeat(duration_width + 1));
```
Also, the code printing the `nerdfont` chars--and in the else branch.
In all these places, you're sending UTF-8 bytes down to the library while the program is in the `C`/`POSIX` locale--which is incorrect.

(As an aside, neither you nor your libraries should print most of these UTF-8 glyphs directly for indicators or the box borders and such. That's what the `ACS_` `curses` constants are there for.)

> I would agree if this bug was about processing regional data as you have outlined, but this is not the case.

Correct. But, regional data is just one aspect. The other use of `setlocale` is to prepare the libraries the program uses for the kind of characters the program will do I/O with. In your case UTF-8. For this, you must call `setlocale`.

Try this test: run `fgrep -r setlocale ...` on what-ever code you have (any language), and count the no. of calls made in the main application vs. no. of calls in the application's libraries. Barring this ncurses-related Rust code, application-called `setlocale` count will >> the library-called `setlocale` count.

hrkfdn:
> But, you are passing raw UTF-8 yourself, too.

That's correct. Fair point, but even if I didn't, the UI would still break (see also https://github.com/gyscos/cursive/issues/494, which does not print any UTF-8 chars on application level). As we have found out now the `setlocale` is already be in place in either `ncurses-backend` or the `pancurses` library, so theoretically it should already work. But it doesn't.

Yes, I could merge this. Maybe I will, as this turns out to be quite the rabbithole. But it remains a bandaid on top of broken code in upstream libraries, such as possibly `ncurses-rs`.

gyscos:
It's hard to test without access to a BSD system, but:
* Does this patch still work if it uses `ncurses::setlocale` instead of `libc::setlocale`?
* Can you `gdb` the example process and check if `ncurses::setlocale` is indeed called by pancurses in the pre-init phase?
* Does this patch still work if the call to `setlocale` is moved to _just before_ initializing `Cursive`?

krytarowski:
This is for the reproducer from https://github.com/gyscos/cursive/issues/494

```
(gdb) b setlocale
Breakpoint 1 at 0x4387b: file /home/kamil/.cargo/registry/src/github.com-1ecc6299db9ec823/ncurses-5.99.0/src/lib.rs, line 1189.
(gdb) r
Starting program: /tmp/cursive_example/target/debug/cursive_example 

Breakpoint 1, ncurses::setlocale (lc=ncurses::constants::LcCategory::all, locale=...)
    at /home/kamil/.cargo/registry/src/github.com-1ecc6299db9ec823/ncurses-5.99.0/src/lib.rs:1189
1189        let c_str = locale.to_c_str();
(gdb) bt
#0  ncurses::setlocale (lc=ncurses::constants::LcCategory::all, locale=...)
    at /home/kamil/.cargo/registry/src/github.com-1ecc6299db9ec823/ncurses-5.99.0/src/lib.rs:1189
#1  0x00000000002354f3 in cursive::backends::curses::n::Backend::init ()
    at /home/kamil/.cargo/registry/src/github.com-1ecc6299db9ec823/cursive-0.15.0/src/backends/curses/n.rs:73
#2  0x00000000002282e1 in core::ops::function::FnOnce::call_once ()
    at /rustc/49cae55760da0a43428eba73abcb659bb70cf2e4/src/libcore/ops/function.rs:232
#3  0x0000000000232faa in cursive_core::cursive::Cursive::try_new (backend_init=0x100000001)
    at /home/kamil/.cargo/registry/src/github.com-1ecc6299db9ec823/cursive_core-0.1.1/src/cursive.rs:107
#4  0x00000000002351de in <cursive_core::cursive::Cursive as cursive::cursive_ext::CursiveExt>::ncurses ()
    at /home/kamil/.cargo/registry/src/github.com-1ecc6299db9ec823/cursive-0.15.0/src/cursive_ext.rs:84
#5  0x00000000002351a2 in <cursive_core::cursive::Cursive as cursive::cursive_ext::CursiveExt>::default ()
    at /home/kamil/.cargo/registry/src/github.com-1ecc6299db9ec823/cursive-0.15.0/src/cursive_ext.rs:74
#6  0x00000000002351fe in cursive::default ()
    at /home/kamil/.cargo/registry/src/github.com-1ecc6299db9ec823/cursive-0.15.0/src/lib.rs:83
#7  0x0000000000222132 in cursive_example::main () at src/main.rs:5
(gdb) frame 0
#0  ncurses::setlocale (lc=ncurses::constants::LcCategory::all, locale=...)
    at /home/kamil/.cargo/registry/src/github.com-1ecc6299db9ec823/ncurses-5.99.0/src/lib.rs:1189
1189        let c_str = locale.to_c_str();
(gdb) list 1180
1175    pub fn scrollok(w: WINDOW, bf: bool) -> i32
1176    { unsafe { ll::scrollok(w, bf as ll::c_bool) } }
1177    
1178    
1179    pub fn scr_restore(filename: &str) -> i32
1180    { unsafe { ll::scr_restore(filename.to_c_str().as_ptr()) } }
1181    
1182    
1183    pub fn scr_set(filename: &str) -> i32
1184    { unsafe { ll::scr_set(filename.to_c_str().as_ptr()) } }
(gdb) 
1185    
1186    pub fn setlocale(lc: LcCategory, locale: &str) -> String
1187    {
1188      unsafe {
1189        let c_str = locale.to_c_str();
1190        let buf = c_str.as_ptr();
1191        let ret = ll::setlocale(lc as libc::c_int, buf);
1192        if ret == ptr::null() {
1193            String::new()
1194        } else {
(gdb) n str
Attempt to use a type name as an expression
(gdb) p str
Attempt to use a type name as an expression
(gdb) p c_str
No symbol 'c_str' in current context
(gdb) n
1190        let buf = c_str.as_ptr();
(gdb) p c_str
$1 = std::ffi::c_str::CString {inner: alloc::boxed::Box<[u8]> {data_ptr: 0x7f7ff7ee4020 <error reading variable>, length: 1}}
(gdb) p *c_str
Attempt to take contents of a non-pointer value.
(gdb) n
1190        let buf = c_str.as_ptr();
(gdb) p buf
$2 = 0x0
(gdb) n
1191        let ret = ll::setlocale(lc as libc::c_int, buf);
(gdb) p buf
$3 = (*mut i8) 0x7f7ff7ee4020 ""
(gdb) step

Breakpoint 1, setlocale (category=0, locale=0x7f7ff7ee4020 "")
    at /usr/src/lib/libc/compat/locale/compat_setlocale1.c:60
60              __mb_len_max_runtime = 1;
(gdb) n
61              return __setlocale(category, locale);
(gdb) step
__setlocale (category=0, name=0x7f7ff7ee4020 "") at /usr/src/lib/libc/locale/setlocale.c:169
169     {
(gdb) n
175             sl = _find_category(category);
(gdb) 
176             if (sl == NULL)
(gdb) 
178             cache = malloc(sizeof(*cache));
(gdb) 
179             if (cache == NULL)
(gdb) 
182             result = (*sl)(name, loc);
(gdb) 
183             _setlocale_cache(loc, cache);
(gdb) p result
$4 = 0x7f7ff6ddd668 <_lc_global_locale+8> "C/C/pl_PL.UTF-8/pl_PL.UTF-8/pl_PL.UTF-8/pl_PL.UTF-8"
(gdb) n
184             return __UNCONST(result);
(gdb) p result
$5 = 0x7f7ff6ddd668 <_lc_global_locale+8> "C/C/pl_PL.UTF-8/pl_PL.UTF-8/pl_PL.UTF-8/pl_PL.UTF-8"
(gdb) n
ncurses::setlocale (lc=ncurses::constants::LcCategory::all, locale=...)
    at /home/kamil/.cargo/registry/src/github.com-1ecc6299db9ec823/ncurses-5.99.0/src/lib.rs:1192
1192        if ret == ptr::null() {
(gdb) n
1197            CStr::from_ptr(ret).to_string_lossy().into_owned()
(gdb) p ret
$6 = (*mut i8) 0x7f7ff6ddd668 <_lc_global_locale+8> "C/C/pl_PL.UTF-8/pl_PL.UTF-8/pl_PL.UTF-8/pl_PL.UTF-8"
(gdb) n
1192        if ret == ptr::null() {
(gdb) 
1199      }
(gdb) 
1200    }
(gdb) 
cursive::backends::curses::n::Backend::init ()
    at /home/kamil/.cargo/registry/src/github.com-1ecc6299db9ec823/cursive-0.15.0/src/backends/curses/n.rs:78
78              ::std::env::set_var("ESCDELAY", "25");
(gdb) 
82              let tty_path = CString::new("/dev/tty").unwrap();
(gdb) c
Continuing.
[Inferior 1 (process 17394) exited normally]
```

gyscos:
So `setlocale` _is_ called, and I assume it's the same `setlocale` as the once called by `libc`.
The question is then how does it differ from calling `libc::setlocale` directly as done in this PR:
* Does the call itself differ? Maybe different arguments? You printed the `buf` being given as an empty string; what is the value of `lc` at this point? And the same when calling `libc::setlocale` with this patch.
* Does the location of the call matter? Does it _have_ to be the first~ish thing a program does, or can it be called later on, as it does right now? (In other words, do we do something we shouldn't between the start of the program and intializing Cursive, where we run `setlocale`?)

krytarowski:
I have figured it out. `setlocale` called by `ncurses::setlocale` is using legacy compat symbol `setlocale` from libc, while libc::setlocale uses the modern one `__setlocale50`.

`char *setlocale(int, const char *) __RENAME(__setlocale50);` 

https://nxr.netbsd.org/xref/src/include/locale.h#90

It's an open question whether the old symbol should still work, but generally Rust should use the newest version always.

gyscos:
Arf indeed, I missed it in libc:
```
    #[cfg_attr(target_os = "netbsd", link_name = "__setlocale50")]
    pub fn setlocale(
        category: ::c_int,
        locale: *const ::c_char,
    ) -> *mut ::c_char;
```

I suppose `ncurses` needs the same sort of annotation to change its linking.

krytarowski:
@gyscos could you please patch rust-libc or ncurses-rs please as a rust programmer?

krytarowski:
@hrkfdn assuming that POSIX idioms are relevant to rust, please add explicit call for `setlocale` in ncspot, even if it is abstracted by ncurses wrapper or a curses implementation.

rajeevvp:
> ...assuming that POSIX idioms are relevant to rust,...

`setlocale()` is not just a POSIX standard, but, a C standard too; which is why `setlocale(LC_ALL, "");`is correct on Windows also.

> ncspot uses a library for the terminal UI to be ncurses/platform agnostic.

These display issues could all be avoided, if you/the library used the `curses`-provided `ACS_*` constants for glyph drawing. Using raw UTF-8 for glyphs even with `setlocale` only works because the program is running inside modern _terminal emulators_. Run `ncspot` or the `cursive` example programs on a real VT-like terminal and you'll get the same result as in the screen-shot, because real terminals don't do UTF-8 at all. Granted, you can't avoid UTF-8 for things like song-titles and such, but, UI-elements can be made much more portable than currently is (if not quite as good-looking).

If you say who uses a DEC VT220 these days, do a search on YouTube. Expect bug reports with unhappy snapshots of real terminals showing the same garbled borders from some of those users. :-)

gyscos:
So:
* `setlocale` is broken on NetBSD for reasons™
* `__setlocale50` is the recommended function to use, and the rust `libc` is already linking to this function, while the `setlocale` from `ncurses-rs` is not NetBSD-aware and just always use the `setlocale` symbol.
* At this point there is little reason for `setlocale` to exist in `ncurses-rs` since `libc` does it, and better, and is already a dependency of `ncurses-rs` anyway.
* I posted a PR to pancurses to stop using ncurses' `setlocale` and switch to `libc`: https://github.com/ihalila/pancurses/pull/78

Re: using ACS in cursive, it's a different discussion but would make it much harder to work with other backends like `BearLibTerminal`, `termion`, `crossterm`, ...
Any non-UTF8 system is explicitly not supported by Cursive, but in practice it shouldn't affect most users - even the linux framebuffer console is UTF8 aware. I would expect people who use actual terminal to be aware of this, and to have some sort of utf8-to-ascii translation layer already to handle this case. Let's not let this affect modern software design.

dricottone:
Speaking as a user of these libraries, please pursue this as an upstream issue. I don't have a NetBSD install, and certainly don't have a non-UTF-8 install--I can't check for these kinds of bugs. At bare minimum, if this sort of hack is necessary for greater portability, it should turn into an upstream documentation patch.

rajeevvp:
> Any non-UTF8 system is explicitly not supported by Cursive, but in practice it shouldn't affect most users - even the linux framebuffer console is UTF8 aware.

Sure, if you make that mandate explicit, then it's OK as far as I'm concerned. But, be aware that, even if the Linux FB console groks UTF-8, other _modern_ console counterparts do not. NetBSD's standard console, `wsvt25` does not. Nor, I think, OpenBSD's terminal console. These only do ASCII, ISO-8859-_n_, KOI-R, ...

For these terminals, you have to use `LC_ALL=C` (or equivalent 8-bits) and only output straight ASCII, KOI-R, ... or `ACS_*` chars.

rajeevvp:
> ...if this sort of hack is necessary for greater portability, ...

That's the point: this is not a **hack**. The proposed patch is doing exactly what the `ISO C`/`POSIX` standards say should be done on locales other than `C`/`POSIX`; and the garbled borders [here](https://github.com/gyscos/cursive/issues/494) are exactly what results from not following that standard.

Now, where it should be done is left floating:
1. I feel `setlocale(LC_ALL, "");` should be done in the main application and the library should use what I have set and not use its own overrides.
2. Or, you could mandate a UTF-8 locale from the user and then the library can assume it's in a UTF-8 locale after calling `setlocale(LC_ALL, "");` and read/write UTF-8 chars, like it does now. But, this way, to me, is just hobbling a program which could display perfectly well (see the `mc`--Midnight Commander--program which works on most glass ttys), with care, on non-UTF-8 consoles too.

krytarowski:
@gyscos Thanks! Please submit a patch to ncurses-rs and switch them to libc `setlocale`+`LC_ALL`. This symbol already leaked into the library and it might be easier at this point to just redirect it to libc.

krytarowski:
While there, another bug:

```
          ld: /tmp/ncspot/target/debug/deps/libtime-166e0e30576a5331.rlib(time-166e0e30576a5331.time.ey6x0hik-cgu.1.rcgu.o): in function `time::tzset':
          /home/kamil/.cargo/registry/src/github.com-1ecc6299db9ec823/time-0.1.43/src/lib.rs:287: warning: warning: reference to compatibility tzset(); include <time.h> for correct reference
```

gyscos:
This looks like an issue for the [time](https://github.com/time-rs/time) crate.
I submitted [an issue](https://github.com/time-rs/time/issues/272) and [a PR](https://github.com/time-rs/time/pull/273). Note though that if merged, it would most likely only affect their 0.2.x branch, while the issue here affects `time-0.1.43`.

krytarowski:
@gyscos great work!

Meanwhile, I wonder what is the way to go to use our native (and original) curses(3) implementation. Should we create netbsd-curses-rs that is similar to ncurses-rs?

The main missing feature today is lack of mouse API, while most other interfaces are interchangeable with ncurses.

gyscos:
If no rust binding currently exist to this library, then you would have to create such a `netbsd-curses-sys` crate (and possibly a "safe" wrapper on top?).
Once this is done and rust can call functions from this library, a new backend for Cursive would have to be written. If the API is similar to ncurses then I assume a lot of code could be shared with the current pan/ncurses backends.

Ramiferous:
Sorry I haven't got a whole lot to add except a screenshot of the patched version which seems to be working as intended. 
This only works with `use_nerdfonts=false` in the `config.toml`

hrkfdn:
I have merged this fix with minor changes until upstream releases are out. Thanks everyone.

voidpin:
@hrkfdn thanks!

PR #263, opened 2020-09-02
Reduce binary size for release profile

r4v3n6101:
Made using this: https://github.com/johnthagen/min-sized-rust


hrkfdn:
Hi, thanks for the PR.

Is there a significant gain? Some of these changes make developing a little harder, e.g. not showing a backtrace. Also optimizing for size over speed may not be in be best interest of everyone, especially since size isn't such a big problem these days.

r4v3n6101:
Ok, let it be in such way. Thanks for response!

Issue #267, opened 2020-09-17
[Request] Add album name to queue/current track + highlight timeout


Ramiferous:
Is there any way to list the Album in the queue ie. [ Artist - Album - Track ]
If I'm listening to multiple albums by the same Artist it's impossible to know which tracks are from which album. Perhaps this could be set in the config file for people who want it?
OR
At the very least, have the current track's album displayed along side the artist and track name under the progress bar.

I also think it would look nice if the selection bar highlighted in red could timeout after 5 or 10 seconds, thus making it look a little cleaner. Again, the timeout delay could be set in the config.yml.

Cheers

jefftherobot:
I see this has been added, and works great. However, can this be configurable? I have instances where my terminal width is small and prefer album column off.

Ramiferous:
> I see this has been added, and works great. However, can this be configurable? I have instances where my terminal width is small and prefer album column off.

I agree, while I usually have ncspot running in a full screen terminal, not everyone does and the album column should be configurable with a toggle setting

hrkfdn:
This is now configurable via `album_column`.

PR #268, opened 2020-09-18
Display album name for playable objects 

mtshrmn:
Adds information about the album in a separate column, here's a picture for example:

image

This also appears in the queue.
Comes to implement #267

### DIFF
--- a/src/album.rs
+++ b/src/album.rs
@@ -159,6 +159,10 @@ impl ListItem for Album {
         format!("{}", self)
     }
 
+    fn display_center(&self) -> String {
+        "".to_string()
+    }
+
     fn display_right(&self, library: Arc<Library>) -> String {
         let saved = if library.is_saved_album(self) {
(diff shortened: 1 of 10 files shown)


hrkfdn:
Nice, thanks. I wonder if something can be done about the spacing. I have tried different things locally, e.g. using only one third as an offset for the center X position, but it always looks a little off 🤔

mtshrmn:
I tried adding a separation line, maybe that would help. Here are some examples with different offsets

2/3 offset:
image
7/11 offset (a bit more than the average of 1/2 and 2/3):
image
7/12 offset (average between 1/2 and 2/3):
image
1/2 offset:
image

hrkfdn:
I actually think 1/2 offset looks just fine, didn't thank of that, so thanks. I took the freedom to update the PR and merge it. Thanks for your contribution!

Issue #269, opened 2020-09-21
[Request] Add option for bitrate


eulerfan271:
Based on other questions, it seems ncspot uses an audio bitrate of 320kb/s.  Higher audio quality great!  However, the higher bitrate uses more cache storage and processing power, and sometimes can't be noticed due to poor headphones, poor hearing, etc.

I thus think it would be useful to have a "bitrate" configuration option (like spotifyd) to lower the bitrate if one wished.

Thanks for this fantastic project!

hrkfdn:
Fixed by #270

PR #270, opened 2020-09-22
Added config option for bitrate

eulerfan271:
Added a config option `bitrate` that allows the bitrate to be set to 96, 160 or 320.

Implements #269


hrkfdn:
Cool, thanks. Could you add the configuration flag to the README? Then it should be fine to merge.

eulerfan271:
Oops, my bad for forgetting to do that the first time.  Added to README, thanks!

PR #271, opened 2020-09-22
Fix last tab to fill entire width correctly

Kemyt:
Condition to check if the tab is the last one was never met. Because of that the last tab wasn't well aligned when terminal width wasn't divisible by the number of tabs.


hrkfdn:
Merged, thanks!

Issue #253, opened 2020-08-20
ncspot exits when deleting empty queue element entry


mtshrmn:
When launching ncspot and entering the queue, prssing d (remove entry from queue) will cause ncspot to exit. 
I don't think this is intentional, that's why I'm submitting an issue.

hrkfdn:
That's a bug alright. Thanks, fixed.

PR #276, opened 2020-09-30
Better queue behavior 

mtshrmn:
- When deleting the last song, update the ui focus (the actual cause for #253).
- Play next song when deleting the currently playing song from queue.
- When skipping to next song the `RepeatTrack` turns into a `RepeatPlaylist`.
- You can now play the expected previous track when shuffle is turned on .


hrkfdn:
Nice, thanks!

PR #277, opened 2020-09-30
feat: add insert command to paste spotify links

TheBolliwood:
- adds `:insert <url>` command to paste open.spotify.com links in ncspot
- use Ctrl-v or use :insert without an argument to paste from clipboard
- both open the selected item either in a modal or in a view (like when using "open current/selected")


hrkfdn:
Merged, thank you!

PR #278, opened 2020-10-01
Play next

### LINKED ISSUE #234: Button to add to top of queue or 'play next'
This would be far more useful in cases where a user is playing a playlist with a large amount of songs and all of them are in the queue before the song they queued up.

mtshrmn:
fixes #234

Currently, the key for that is `n` as in play [n]ext. 
In my fork, there's a vim like searching with `n` and `N`. I would like to create a pr for that as well, and if it will be merged, the keybindings would be in conflict. I solved it by:

`p` - go to currently playing
`.` - play next (like the `>` but without pressing shift)
`n` - search next

What do you think the keybindings should be?


src/playable.rs
```diff
@@ -88,6 +88,10 @@ impl ListItem for Playable {
         self.as_listitem().play(queue)
     }
 
+    fn play_next(&mut self, queue: Arc<Queue>) {
+        self.queue(queue)
```
hrkfdn, line 92: Is there a reason this does not call `play_next` on the `ListItem` but `queue` instead?

mtshrmn: No idea how that got there, there's no reason for that. Fixing it now.

hrkfdn:
Merged, thank you. I'll gladly take a look at the PR and Vim search sounds nice, but as keybindings are very subjective I'd prefer not to change the existing bindings too much, especially because I'm not a Vim user myself. Just a little disclaimer before you put in the effort and energy :smile:

Your proposal sounds reasonable, though.

mtshrmn:
Well, the code is already written so too late 😅
I'll just let you decide on the keybindings, I can always change them locally in the config file.

PR #285, opened 2020-10-11
Replace "directories" crate with "platform_dirs"

marsam:
The `directories` create is no longer maintained.
Also it uses `~/Library/Preferences` on MacOS (https://github.com/dirs-dev/directories-rs/issues/62), but Apple advises against [creating files in `Preferences` sufolder](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW1):
>This directory contains app-specific preference files. You should not create files in this directory yourself. Instead, use the NSUserDefaults class or CFPreferences API to get and set preference values for your app.In iOS, the contents of this directory are backed up by iTunes and iCloud.

[`platform-dirs`](https://github.com/cjbassi/platform-dirs-rs) allows to use `~/.config` on MacOS.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -806,16 +806,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "directories"
-version = "2.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "551a778172a450d7fc12e629ca3b0428d00f6afa9a43da1b630d54604e97371c"
-dependencies = [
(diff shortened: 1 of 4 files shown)


hrkfdn:
Thanks, I guess with the CLI context switching over to XDG paths is reasonable for macOS. Merged!

PR #298, opened 2020-10-20
Increase/Decrease volume by 5

ChaboCode:
I choosed `[` and `]` because are near to `-` and `+`.
I tried to use a keybinding with `-` and `+`, but `Ctrl` was used by the terminal and `Alt` wasn't recognized.

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -99,8 +99,8 @@ pub enum Command {
     Delete,
     Focus(String),
     Seek(SeekDirection),
-    VolumeUp,
-    VolumeDown,
+    VolumeUp(u16),
+    VolumeDown(u16),
     Repeat(Option<RepeatSetting>),
     Shuffle(Option<bool>),
     Share(TargetMode),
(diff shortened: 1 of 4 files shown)


src/command.rs
```diff
@@ -344,8 +344,8 @@ pub fn parse(input: &str) -> Option<Command> {
                 _ => Command::Save,
             })
             .or(Some(Command::Save)),
-        "volup" => Some(Command::VolumeUp),
-        "voldown" => Some(Command::VolumeDown),
+        "volup" => Some(Command::VolumeUp(1)),
+        "voldown" => Some(Command::VolumeDown(1)),
```
hrkfdn, line 348: This will always use `1` for the amount when parsing the bindings

hrkfdn (changes requested):
Thanks for the PR. There is one problem with the parsing of custom `volume N` bindings where `N` is basically ignored and a static value of `1` is used for the amount. Could you please adapt this?

ChaboCode:
Can you label this `hacktoberfest-accepted` please?

hrkfdn:
Merged, thanks.

PR #308, opened 2020-10-27
Changes to documentation

ChaboCode:
I think this is important, so the user can set custom bindings to decrease the volume by 50%, or things like that.


hrkfdn:
Hey @ChaboCode, I appreciate the PR, but I believe it is implied by the examples already provided, so I'd close this PR for now.

Issue #197, opened 2020-05-14
Feature request: open external links


BachoSeven:
It would be awesome to be able to open external links to songs/playlists in ncspot!

mxdevmanuel:
I think this is already requested in #155

Bettehem:
You can do this via MPRIS already.
For example with playerctl you can run
```
playerctl -p ncspot open <your_link_goes_here>
```

BachoSeven:
@Bettehem thanks for the tip, I did not know about `playerctl`, however:
- this solution only works when ncspot is already open,
- even when it is, it does not work for me, meaning the command is executed but nothing happens in the player.

I would be happy with playerctl if I was able to, say, write a script that opens dmenu and lets me paste a link to open in ncspot

BachoSeven:
Okay now I figured it out: in order for it to work through playerctl, you need the following format for the URL:
(Example)
`https://open.spotify.com/track/2WTAUN77y319PP2TKdbxgw` -->> 
 `spotify:track:2WTAUN77y319PP2TKdbxgw`

Then, `playerctl -p ncspot open URL` will work as mentioned. This is what I was looking for, so I will close the issue.

Issue #155, opened 2020-03-18
Ability to open shared tracks


ltning:
I'm getting links from colleagues and friends of the form 'https://open.spotify.com/album/3mpabcdef382e79yhif34e' (random non-working sample), I would very much like to be able to open those in ncspot. Either through a "open URL" or passing it on a pipe or whatever.

Any takers? :)

hrkfdn:
Possibly related to https://github.com/hrkfdn/ncspot/issues/97

Bettehem:
I'll look into this. Shouldn't be too hard to add so you can use the MPRIS openUri function to open a shared link too.
In the meanwhile if you want to open a shared track in ncspot, you first have to convert the link to a URI:
`https://open.spotify.com/album/3mpabcdef382e79yhif34e` -> `spotify:album:3mpabcdef382e79yhif34e`
Then use something like [playerctl](https://github.com/altdesktop/playerctl) to open it by running `playerctl -p ncspot open spotify:album:3mpabcdef382e79yhif34e`

hrkfdn:
This is now implemented via MPRIS and CTRL-V shortcut. Closing issue.

PR #314, opened 2020-11-08
Spotify URL support for MPRIS OpenUri function

Bettehem:
Added support for spotify urls to MPRIS OpenUri function.
Previously when using something like playerctl to open a link sent by a friend, you would have to manually convert the spotify URL into a spotify URI for it to work. Now it works both ways.
I also fixed mpris:trackid returning the id in a weird format. This broke my music bot and I thought there was something wrong with my code but turned out the id formatting was off here.

Edit: This should also better solve [#197 ](https://github.com/hrkfdn/ncspot/issues/197) and [#155 ](https://github.com/hrkfdn/ncspot/issues/155)


hrkfdn:
Merged with some changes, thank you!

Bettehem:
Nice, thanks!

PR #322, opened 2020-11-17
Change to correct trackid

r4v3n6101:
Seems that one of commit from #252 was lost so trying to do something with MPRIS via `playerctl` crashes because there's wrong `mpris:trackid` in the code.


hrkfdn:
Merged, thanks

Bettehem:
How is this correct? `mpris:trackid` should be in the format of `spotify:<type>:<id>` aka Spotify URI.
This is also how the official Spotify client works. it always returns the current tracks URI in `mpris:trackid`

PR #325, opened 2020-11-20
fix mpris get volume; added mpris set volume

### LINKED ISSUE #315: MPRIS set volume support
I have keybinds that use `playerctl` to adjust the volume of the active player (which already works great for cmus and mpv), so it would be great if we could add volume MPRIS support to ncspot. I have not done any rust development before, but it looks like it would be as simple as adding a single function and registering it with dbus, and there are plenty of similar functions in the existing code base. 

Let me know if this is a welcomed change and I can take a crack at a pull request.

benjaminBrownlee:
The MPRIS function to get volume was hardcoded to always return 100%; it now returns the actual volume. Also added MRIS function to set the volume. Both appear to be working on my device with playerctl.

resolves #315

### DIFF
--- a/src/mpris.rs
+++ b/src/mpris.rs
@@ -19,7 +19,7 @@ use crate::playable::Playable;
 use crate::playlist::Playlist;
 use crate::queue::{Queue, RepeatSetting};
 use crate::show::Show;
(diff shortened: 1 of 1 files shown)


hrkfdn:
Merged, thank you!

Issue #189, opened 2020-04-29
Allow users to change sorting order in ListView


hrkfdn:
Users should be able to change the sorting order in `ListView` based views. Ideally, the sorting order can be set depending on the available columns and is persisted per view.

This will help with:
- #140
- #188

PR #328, opened 2020-11-21
Add support for sorting playlist tracks

Qluxzz:
This PR adds support for sorting tracks in a playlist, the function for sorting is generic but the command is only valid for playlists for now.

#189

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -44,6 +44,23 @@ impl Default for MoveAmount {
 
 #[derive(Display, Clone, Serialize, Deserialize, Debug)]
 #[strum(serialize_all = "lowercase")]
+pub enum SortKey {
+    Title,
+    Duration,
+    Artist,
+    Album,
+    Added,
+}
+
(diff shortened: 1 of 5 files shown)


src/command.rs
```diff
@@ -173,6 +191,7 @@ impl fmt::Display for Command {
             Command::ReloadConfig => "reload".to_string(),
             Command::Insert(_) => "insert".to_string(),
             Command::NewPlaylist(name) => format!("new playlist {}", name),
+            Command::Sort(key, direction) => format!("sort playlist by {} {}", key, direction),
```
hrkfdn, line 194: These should represent the command in a way it can be bound to a key or entered within the console, so I'd assume it's a little more like this:

```suggestion
            Command::Sort(key, direction) => format!("sort {} {}", key, direction),
```

hrkfdn (changes requested):
Thanks, looks good. One minor change is needed in the `fmt::Display` for `Sort` commands.

hrkfdn:
Merged, thanks!

PR #329, opened 2020-11-22
Only show playlists where the user is allowed to add tracks when trying to add tracks

Qluxzz:
Currently all playlists are showed when trying to add a track, this fixes so only playlists where the user is allowed to add tracks are shown.

### DIFF
--- a/src/ui/contextmenu.rs
+++ b/src/ui/contextmenu.rs
@@ -38,10 +38,10 @@ impl ContextMenu {
         track: Track,
     ) -> Modal<Dialog> {
         let mut list_select: SelectView<Playlist> = SelectView::new().autojump();
-        let current_user_id = spotify.current_user().unwrap().id;
+        let current_user_id = library.user_id.as_ref().unwrap();
 
         for list in library.items().iter() {
-            if current_user_id == list.owner_id || list.collaborative {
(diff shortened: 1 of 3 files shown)


src/ui/contextmenu.rs
```diff
@@ -38,9 +38,12 @@ impl ContextMenu {
         track: Track,
     ) -> Modal<Dialog> {
         let mut list_select: SelectView<Playlist> = SelectView::new().autojump();
+        let current_user_id = spotify.current_user().unwrap().id;
```
Qluxzz, line 41: Can I be sure that the current_user always is available here or should I handle this differently?

hrkfdn: Should be fine. The UI doesn't show if ncspot can't create a valid session. Maybe it could be cached within Spotify as it's unlikely to change during runtime to avoid repeated API requests.

Qluxzz: I saw the library struct has a private field with the user_id, maybe making that field public and using it is a better strategy?

hrkfdn: Yeah, that would be one approach.

hrkfdn:
Merged, thanks!

PR #335, opened 2020-11-27
Add hjkl controls for AddToPlaylistMenu

### LINKED ISSUE #334: cannot navigate with vim keybinds in "Add to playlist" menu
I noticed that I can navigate with hjkl everywhere but in the "Add to playlist" menu. Running ncspot 0.2.4.

JojiiOfficial:
Fixes #334

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -2,9 +2,6 @@ use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Duration;
 
-use crate::command::{
-    parse, Command, GotoMode, JumpMode, MoveAmount, MoveMode, SeekDirection, ShiftMode, TargetMode,
-};
 use crate::config::Config;
 use crate::library::Library;
 use crate::queue::{Queue, RepeatSetting};
@@ -14,6 +11,13 @@ use crate::ui::contextmenu::ContextMenu;
 use crate::ui::help::HelpView;
 use crate::ui::layout::Layout;
 use crate::UserData;
+use crate::{
+    command::{
+        parse, Command, GotoMode, JumpMode, MoveAmount, MoveMode, SeekDirection, ShiftMode,
+        TargetMode,
+    },
+    ui::contextmenu::AddToPlaylistMenu,
+};
 use cursive::event::{Event, Key};
 use cursive::traits::View;
 use cursive::Cursive;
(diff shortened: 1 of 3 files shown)


hrkfdn:
Thanks! I just tested this but it seems to be clashing with cursive's autojump feature on my machine. Are you not experiencing this? For instance, when I repeatedly press `l` in the `AddToPlaylistMenu` it will jump between the playlists starting with "l".

JojiiOfficial:
I probably don't have enough playlists to experience a collision there. Whats the best approach for fixing this? Disabling one of both features?

hrkfdn:
I guess so, the question is which? 😄 I suppose the Vim users would like their bindings but I like the autojump feature quite a lot, especially if users have a lot of playlists. Thoughts anyone?

JojiiOfficial:
An option would be to allow the user to choose which one. I'm a vim user so I prefer the vim keybindings but I'm sure there are many people not being that much into vim. Maybe a config option `keystile` or `bindingstyle` which can be set to either `vim` or `autojump`? That would make it a bit confusing since its a bit of vim either way. Probably a proper trait implementation for autojump/vim bindings would be more clean but I don't know by heart how easy that'd be.

Why does the vim keybindings don't interfere with the context menu though?

Bettehem:
I have an idea. What if it would work by default with normal vim bindings and then by pressing `/` to enter search/find mode and then pressing `l` or `j` etc. would jump to playlists starting with that letter. This could possibly be implemented in any dialog/popup window so if you quickly want to jump to a specific option in the menu, you'd just need to press `/` and then the letter the option begins with.
So for example if you want to jump to "Add to playlist" in a dialog menu just press `/` followed by `a`
What do you think of this?

invakid404:
The idea @Bettehem gave sounds really reasonable to me.

hrkfdn:
I have disabled autojump for now in favor of this PR. Autojump is not documented and key bindings should definitely be supported in the context menu. If a playlist search function or similar within the context menu is required it can still be implemented separately.

Thanks for the PR @JojiiOfficial.

BachoSeven:
@hrkfdn I just updated to `master`, but I can't confirm that the issue is solved.

Pressing `j` while in the "Add to Playlist" context still cycles through the playlists whose name starts with `j` via autojump. 

`ncspot -V` gives 0.2.5

hrkfdn:
Mhh, odd. Can't really reproduce this. I have playlists starting with `J` and `K` but the selection moves down/up here.

Bettehem:
I can't reproduce the problem either. For me it works as it should, going down with `j` and up with `k`

BachoSeven:
Oh ok. I'll try to recompile it, maybe something went wrong while updating

BachoSeven:
I forgot to report it, but I still experience the issue, even on `0.4.0`.

PR #347, opened 2020-12-14
Allow primary selection for sharing

ninewise:
Adds a feature flag to use the primary selection instead of the clipboard. Only works on Linux.


hrkfdn:
Merged, thank you!

PR #363, opened 2021-01-04
Enable mouse scrolling in TabViews

algon-320:
Before this patch, we couldn't scroll ListViews contained by TabView with mouse.
This is because the default implementation of `cursive::traits::View::on_event` ignores any event.
This PR fixes the problem by forwarding events properly.


hrkfdn:
Merged, thanks a lot!

PR #369, opened 2021-01-07
Expose gapless playback config option

Qluxzz:


hrkfdn:
Thanks for the PR. Have you tested this? I don't know if this works as ncspot does not use the SPIRC/Spotify remote player that is used in librespot/spotifyd, as ncspot has its own queue.

Qluxzz:
Yes, I have tested it with Pink Floyds The Dark Side of the Moon since it has tracks that blends together and it worked great, no more silence between tracks!

hrkfdn:
Cool, thanks for confirming!

deed02392:
Wow, this works really well. We might want to consider this default enabled.

PR #370, opened 2021-01-07
Fix not setting correct country/market

Qluxzz:
Previously the country was not set at all. This had the behaviour that when country was required it was defaulted to the US, and when not required it would get all entries ignoring your current market. This had some weird implications especially when looking at an artists albums
image

After setting the correct country the list now looks like what I see in the desktop client:

image

### DIFF
--- a/src/spotify.rs
+++ b/src/spotify.rs
@@ -12,7 +12,7 @@ use librespot_playback::config::Bitrate;
 use librespot_playback::mixer::Mixer;
 use librespot_playback::player::{Player, PlayerEvent as LibrespotPlayerEvent};
 
-use rspotify::blocking::client::ApiError;
+use rspotify::{blocking::client::ApiError, senum::Country};
 use rspotify::blocking::client::Spotify as SpotifyAPI;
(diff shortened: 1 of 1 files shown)


hrkfdn:
I guess this is the way to go for now until the behaviour is changed upstream: https://github.com/ramsayleung/rspotify/issues/28

This can be determined automatically by the Spotify API, but rspotify falls back to `US` if no value is set: https://docs.rs/rspotify/0.10.0/src/rspotify/client.rs.html#350

> - If market is set to from_token, and a valid access token is specified in the request header, only content playable in the country associated with the user account, is returned.

Merged, thanks!

Qluxzz:
> This can be determined automatically by the Spotify API, but rspotify falls back to `US` if no value is set: https://docs.rs/rspotify/0.10.0/src/rspotify/client.rs.html#350

I see what you mean, weird that the current fallback is the US instead of "from_token" since it would solve every problem.

Seems like rspotify is active enough so I will see if I can't create a PR to rspotify with the changes required.

PR #375, opened 2021-01-09
Add select artist dialog when multiple artists

Qluxzz:
Currently when using "show artist" or "goto artist" and multiple artists are present, only the first artist is shown. This PR adds a new select dialog where the user gets to select which artist to go to.

The dialog is only shown when multiple artists are present, otherwise it goes directly to the artist.

image


hrkfdn:
Looks good! Thank you, merged.

PR #382, opened 2021-01-12
patch enumset dependency to work around build failure

mworthley:
While installing via Cargo, I received `cannot find export in syn` due to an issue with enumset. I found a workaround (link below) and thought I would share since I'm not sure when their permanent fix will be available. Tested locally and build now succeeds.

https://github.com/Lymia/enumset/pull/18#issuecomment-756358826

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -940,8 +940,7 @@ dependencies = [
 [[package]]
 name = "enumset_derive"
 version = "0.5.0"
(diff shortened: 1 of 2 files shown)


hrkfdn:
Thanks for the PR. This will fix the build, but I won't be able to publish new releases to crates.io because git dependencies are not allowed. I'll have to find another approach.

PR #392, opened 2021-01-15
Add support for open.spotify.com links

Qluxzz:
Adds support for open.spotify.com link such as https://open.spotify.com/track/4uLU6hMCjMI75M1A2tKUQC

Same behaviour as with Spotify URI:s, paste them in the searchbar and press enter.


Qluxzz:
Oh wow, I just now saw that you'd already implemented similar functionality here:
https://github.com/hrkfdn/ncspot/blob/d9d6d684ea04b0ab1eadee62f4e132fd8f4523ed/src/ui/listview.rs#L644-L700

hrkfdn:
Right, this was implemented in #277. I think this should be merged somehow.

Qluxzz:
The insert command might need to be addressed in a separate PR.
Some media types only result in a box with share being the only example.
image

hrkfdn:
Very nice, thank you for reiterating.

PR #396, opened 2021-01-16
Fix multiple tracks in queue being shown as playing

Qluxzz:
Before this fix, when adding duplicates to the queue, all duplicates would be shown as playing:
Example of multiple tracks being marked as playing

After, only the currently playing track is shown as playing.
image


Qluxzz:
This fixes the queue view, but if playing a playlist and navigating to the album, the currently playing track on the album will not be shown as playing as before. However that is how it's currently shown in the client.

hrkfdn:
Thank you, merged!

PR #397, opened 2021-01-16
Return to search when viewing search results

### LINKED ISSUE #390: Return to search when viewing search results
With the changes that came from separating the search and search results in commit https://github.com/hrkfdn/ncspot/commit/18dc6c6bf84a5529c200376020eda48bd0ea7557 searches now opens a new view with the search results.

If you navigate a bit through the search results and navigate to albums and artists it takes multiple keypresses to return to the search view.

My solution is if you're viewing the search results and press to focus the search view again, the search results are closed and you're returned to the search view.

@hrkfdn, could you maybe give me some hints on how to be able to pop the view or otherwise programmatically close the search results?

I've tried reusing the same logic as pushing the view from the on_submit handler in the on_command handler but from that I can see pop_view is never called so I guess call_on_name doesn't work there.
```Rust
fn on_command(&mut self, s: &mut Cursive, cmd: &Command) -> Result<CommandResult, String> {
    if let Command::Focus(view) = cmd {
        if view == &"search" && self.showing_results {
            s.call_on_name("main", move |v: &mut Layout| v.pop_view());
            self.showing_results = false;
        }

        self.edit_focused = true;
        self.clear();
        return Ok(CommandResult::Consumed(None));
    }

    Ok(CommandResult::Ignored)
}
```

Qluxzz:
With the changes that came from separating the search and search results in commit 18dc6c6 searches now opens a new view with the search results.

If you navigate a bit through the search results and navigate to albums and artists it takes multiple keypresses to return to the search view.

My solution is if you're viewing the search results and press to focus the search view again, the search results are closed and you're returned to the search view.

Fixes #390

### DIFF
--- a/src/ui/layout.rs
+++ b/src/ui/layout.rs
@@ -318,6 +318,15 @@ impl ViewExt for Layout {
     fn on_command(&mut self, s: &mut Cursive, cmd: &Command) -> Result<CommandResult, String> {
         match cmd {
             Command::Focus(view) => {
+                // Clear search results and return to search bar
+                // If trying to focus search screen while already on it
+                let search_view_name = "search";
+                if view == search_view_name && self.focus == Some(search_view_name.into()) {
+                    if let Some(stack) = self.stack.get_mut(search_view_name) {
+                        stack.clear();
+                    }
+                }
+
                 if self.screens.keys().any(|k| k == view) {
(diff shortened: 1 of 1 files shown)


hrkfdn:
Merged, thank you. I wonder if we should do this for all screens, but this can still be changed later if desired.

PR #399, opened 2021-01-18
Restart queue if pressing play when stopped

Qluxzz:
Currently when all tracks in the queue have been played, and the playback has stopped, the play/pause button does nothing.

This PR copies the behaviour from the desktop client, with the queue restarting from the top and starting to play again if play/pause is pressed.

It also fixes a bug where if adding a track after playback had stopped, play/pause did nothing but next track would advance and play the new track in the queue. Now when pressing play/pause it plays the next track in the queue as expected.

### DIFF
--- a/src/mpris.rs
+++ b/src/mpris.rs
@@ -400,9 +400,9 @@ fn run_dbus_server(
         });
 
     let method_playpause = {
-        let spotify = spotify.clone();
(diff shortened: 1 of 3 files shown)


hrkfdn:
Looks reasonable, thank you!

PR #401, opened 2021-01-19
Update to cursive 0.16

gyscos:


hrkfdn:
Thanks for the PR and for Cursive :)

PR #403, opened 2021-01-19
Don't handle FinishedTrack as Stopped

### LINKED ISSUE #395: [mpris] Play State bouncing
**Version**: 0.3.0

## What is seen

When playing tracks, the MPRIS play state bounces when changing tracks.  What I mean is that when a track finishes and the next track starts, MPRIS reports the player `stopped` and then immediately `playing` after.

I see this by using the [`playerctl`](https://github.com/altdesktop/playerctl) MPRIS helper in monitor mode (`playerctl -a -F metadata -f '{{lc(status)}} - {{playerName}}'`) and when playing through a playlist the output looks like:
```
playing - ncspot
stopped - ncspot
playing - ncspot
stopped - ncspot
playing - ncspot
```

## What I expect

When the playing starts, a `playing` status is broadcast and then `stopped` is only broadcast when the queue is empty (i.e. there is no next track to play).  This is the behavior in other music players (this is playing through a three track playlist):
```
playing - cmus
stopped - cmus
```

Qluxzz:
Currently FinishedTrack and Stopped both returned "Stopped" leading mpris to report stopped after every track and then playing again when the new track starts.

Not sure if I am allowed to not return a value when we don't have a valid PlaybackState value. The [docs](https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Property:PlaybackStatussays ) for the PlaybackState states "When this property changes, the org.freedesktop.DBus.Properties.PropertiesChanged signal is emitted with the new value." 

but I'm not very knowledgeable with DBus. 

I've not encountered any weird behaviour using the same debugging technique described in #395.

Fixes #395

### DIFF
--- a/src/mpris.rs
+++ b/src/mpris.rs
@@ -28,13 +28,13 @@ type Metadata = HashMap<String, Variant<Box<dyn RefArg>>>;
 
 struct MprisState(String, Option<Playable>);
 
-fn get_playbackstatus(spotify: Arc<Spotify>) -> Option<String> {
+fn get_playbackstatus(spotify: Arc<Spotify>) -> String {
     match spotify.get_current_status() {
-        PlayerEvent::Playing => Some("Playing".into()),
-        PlayerEvent::Paused => Some("Paused".into()),
-        PlayerEvent::Stopped => Some("Stopped".into()),
-        _ => None,
+        PlayerEvent::Playing | PlayerEvent::FinishedTrack => "Playing",
+        PlayerEvent::Paused => "Paused",
+        _ => "Stopped",
     }
+    .to_string()
 }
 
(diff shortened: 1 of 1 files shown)


hrkfdn:
Thanks for tackling this. I took a look to see what happens if we just add nothing with `gdbus introspect --session --dest org.mpris.MediaPlayer2.ncspot --object-path '/org/mpris/MediaPlayer2' --only-properties` and in those cases the value seems to be empty, which it seems would violate the MPRIS specs. I'd suggest to just keep the state at `Playing` when `FinishedTrack` was emitted, as it should be followed by a `Stopped` event. I will commit the changes in a second.

PR #405, opened 2021-01-23
Add simple debian-packaging with cargo-deb

treibholz:
This adds simple debian-packaging with cargo-deb.

I have seen this on https://github.com/Spotifyd/spotifyd, but I like ncspot much more, so I added it to the Cargo.toml and updated the Readme.md how to build a debian-package.


hrkfdn:
Thanks a lot, merged!

PR #428, opened 2021-02-13
modified library.rs to sort after saving an album

jmmerchant:
Hello.  I just started using ncspot and found that the list of albums in the library were not sorted after saving a new one (say from the list of albums in some search results).  Is this a design decision?  If so, please ignore this.  If not, here's a very quick and dirty fix.


hrkfdn:
Hey, this is a bug indeed, so thanks for your PR. I would have merged this already, but I'm currently working on preserving sorting order for views, playlists, etc. I'm hoping to resolve the issue you reported with that change as well.

PR #437, opened 2021-02-21
LoopStatus is now editable from MPRIS

r4v3n6101:
Make a small useful thing allows to set loop status (one track, one playlist or none of them) using the MPRIS.


hrkfdn:
Merged, thanks!

PR #451, opened 2021-02-26
Implement cover drawing as optional feature

KoffeinFlummi:
I implemented this ages ago, including support for both W3M and Überzug. @hrkfdn I think at the time you mentioned that you thought this was beyond the scope of the project.

I have recently refactored this feature completely so I could finally update my local fork to keep up with the upstream (Great work since then btw). I have thrown out W3M support because that was a complete mess and required X11 packages and other unsavory things. It also required constant redrawing, while Überzug is much easier to work with.

This is an optional feature, so when compiling without it, the only change is the addition of an `on_leave` method for views. Without the additional complexity of W3M, it should be fairly maintenance-free, and it's a feature I can't live without anymore.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1474,6 +1474,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ioctl-rs"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


hrkfdn:
Thanks, I'll gladly merge it!

PR #466, opened 2021-03-15
Refactor needs_download and rename library.items to better convey usage

Qluxzz:
Use .find() instead of for each loop to find playlist.

Just thought it made it clearer than before.

Saw that library had a function called .items() which only returned the playlists so I reused it in needs_download and renamed it to .playlists() since that is what it returns.

### DIFF
--- a/src/library.rs
+++ b/src/library.rs
@@ -60,10 +60,8 @@ impl Library {
         library
     }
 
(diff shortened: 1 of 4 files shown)


hrkfdn:
Reasonable changes, thank you!

PR #467, opened 2021-03-15
Refactor all .filter.map chains to .filter_map

Qluxzz:
Combines all .filter.map chains with a single call to .filter_map instead.

`.filter` yields if closure returns `true`, `.filter_map` yields if closure returns `Some()`


hrkfdn:
Much nicer, thanks!

Issue #485, opened 2021-04-10
Panic loading default user state configuration


chrhasse:
**Describe the bug**
Panics immediately on launch when loading the default userstate configuration, either due to an old, no longer valid file, or no file at all.
```
thread 'main' panicked at 'could not load user state: "Failed serializing value: values must be emitted before tables"', src/config.rs:113:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
If a valid `userstate.toml` exists the program will function properly.

**To Reproduce**
1. Build ncspot as of https://github.com/hrkfdn/ncspot/commit/cea5228245d7bc56bd07aab53547881dac03063f
2. Move or delete ~/.config/ncspot/userstate.toml
3. Launch ncspot
4. Observe panic

**Expected behavior**
Program finishes launching.

**System (please complete the following information):**
 - OS: Archlinux
 - Terminal: Termite
 - Version: master
 - Installed from: AUR `ncspot-git`

**Backtrace/Debug log**
```
    Finished dev [unoptimized + debuginfo] target(s) in 8.63s
     Running `target/debug/ncspot`
thread 'main' panicked at 'could not load user state: "Failed serializing value: values must be emitted before tables"', src/config.rs:113:18
stack backtrace:
   0:     0x55b15480b9f0 - std::backtrace_rs::backtrace::libunwind::trace::h5e9d00f0cdf4f57e
                               at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x55b15480b9f0 - std::backtrace_rs::backtrace::trace_unsynchronized::hd5302bd66215dab9
                               at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x55b15480b9f0 - std::sys_common::backtrace::_print_fmt::ha0237cd11a34e2bf
                               at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x55b15480b9f0 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h171d4c10df1a98ee
                               at /rustc/2fd73fabe469357a12c2c974c140f67e7cdd76d0/library/std/src/sys_common/backtrace.rs:46:22
[...]

PR #486, opened 2021-04-10
Allow creation of default userstate.toml file

chrhasse:
Serializing an empty `Vec<Playable>` causes serde to panic as described
in #485 This commit tells serde not to serialize `queue` if it is an
empty `Vec`.


hrkfdn:
Good catch, thanks!

hrkfdn:
The userstate should probably be stored in a binary format to avoid problems like these. Considering the amount and rising complexity of data structures, TOML may not be the best choice.

PR #488, opened 2021-04-12
Saving album

r4v3n6101:
Don't save track when album does as it's default behavior of Spotify's official app.


hrkfdn:
Makes sense, thank you!

PR #492, opened 2021-04-15
Document authentication and stored credentials.

jez:
Thanks for the great project!

I tend to be a bit cagey about giving my username and password to random CLI scripts, and I wanted to make sure that my password was not persisted to disk. Glad to see that it isn't, so I figured I'd document (my take on) how it works.


hrkfdn:
Looks good, thanks!

PR #493, opened 2021-04-17
Add rounding to volume percentage display

Kemyt:
Shown percentage is sometimes lower than actual because float number is cast to u16 without rounding, so for example 4.99_f64 becomes 4_u16. It's seen for example when pressing "volup 5" when volume is at 0%. When you do that the bar displays 4%.
image


hrkfdn:
Merged, thanks!

PR #517, opened 2021-05-06
Make behavior of Shift modifier consistent

### LINKED ISSUE #516: Inconsistent use of Shift in default keybindings
In the default keybindings, the use of `Shift` as a modifier is inconsistent: `o` opens the context menu for the currently selected item while `Shift+o` opens the menu for the currently playing item, whereas the behavior is reversed for `x`, `x` copies a link to the _currently playing_ item while `Shift+x` copies a link for the _currently selected_ item.

Based on #75, it seems like the intention is for the `Shift` modifier to mean currently playing, so the default behavior for `x` seems like it should be inverted.

rbuch:
Previously, Shift+o opened the menu for the currently playing item
while Shift+x copied a link to the currently selected item. Now Shift
operates on the currently playing item in both cases.

Fixes #516.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,8 +133,8 @@ depending on your desktop environment settings. Have a look at the
 * `r` to toggle repeat mode
 * `z` to toggle shuffle playback
 * `q` quits ncspot
-* `x` copies a sharable URL of the song to the system clipboard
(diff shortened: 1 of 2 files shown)


hrkfdn:
Makes sense, thanks!

PR #525, opened 2021-05-14
Explicitly set default normalisation threshold for librespot

### LINKED ISSUE #522: Highly distorted audio when audio normalization is true
**Describe the bug**
I recently migrated from pulseaudio to pipewire and since then, audio in ncspot sounds very loud and distorted, even when in-app volume is set to 0%.
EDIT: Seems like the problem involves audio normalization and an update to librespot, as indicated on comments below.

**To Reproduce**
1. Uninstall pulseaudio
2. Install pipewire and pipewire-pulse
3. Open ncspot and play any song

**Expected behavior**
Supposedly it shouldn't sound different than if played with pulseaudio.

 - OS: Arch Linux
 - Terminal: Termite
 - Version: 0.7.0-1
 - Installed from: AUR

### LINKED ISSUE #552: freeze after suspend/resome
Kind of a continuation of https://github.com/hrkfdn/ncspot/issues/536

`master` version (`677a8e0`) was working fine for a while, but today it decided to freeze after resuming from sleep and finishing the buffered track. Looks like it got stuck trying to load next track, UI is completely unresponsive. Stack trace from gdb:
```
#0  0x00007f9759529e29 in syscall () from /lib64/libc.so.6
#1  0x0000562ba9d5edea in std::thread::park ()
#2  0x0000562ba939a8d3 in std::thread::local::LocalKey<T>::with ()
#3  0x0000562ba91fa926 in futures_executor::local_pool::block_on ()
#4  0x0000562ba94c6746 in ncspot::spotify::Spotify::refresh_token ()
#5  0x0000562ba94c593b in ncspot::spotify::Spotify::start_worker ()
#6  0x0000562ba92dcf64 in ncspot::main::_$u7b$$u7b$closure$u7d$$u7d$::hf66b629480604a11 ()
#7  0x0000562ba93b07d3 in tokio::runtime::enter::Enter::block_on ()
#8  0x0000562ba926f265 in tokio::runtime::Runtime::block_on ()
#9  0x0000562ba9388ff2 in ncspot::main ()
#10 0x0000562ba929ae53 in std::sys_common::backtrace::__rust_begin_short_backtrace ()
#11 0x0000562ba94b8539 in std::rt::lang_start::_$u7b$$u7b$closure$u7d$$u7d$::h30f1501b7a390c8a ()
#12 0x0000562ba9d71677 in std::rt::lang_start_internal ()
#13 0x0000562ba93890b2 in main ()
```

I don't think this is easily reproducible

rbuch:
The normalisation threshold should be in ratio space, but the default value is specified in dB space, so using it as the default implicitly without passing it through the `db_to_ratio` function causes distortion and clipping when normalisation is enabled.

Fixes #522.

Ideally this would be done internally in `librespot`, so revert this is this logic is added there (see https://github.com/librespot-org/librespot/issues/745).

### DIFF
--- a/src/spotify.rs
+++ b/src/spotify.rs
@@ -4,6 +4,7 @@ use librespot_core::config::SessionConfig;
 use librespot_core::session::Session;
 use librespot_core::session::SessionError;
 use librespot_playback::config::PlayerConfig;
+use librespot_playback::player::NormalisationData;
 use log::{debug, error, info};
 
 use librespot_playback::audio_backend;
@@ -190,6 +191,9 @@ impl Spotify {
             bitrate: bitrate.unwrap_or(Bitrate::Bitrate320),
             normalisation: cfg.values().volnorm.unwrap_or(false),
             normalisation_pregain: cfg.values().volnorm_pregain.unwrap_or(0.0),
+            normalisation_threshold: NormalisationData::db_to_ratio(
+                PlayerConfig::default().normalisation_threshold,
+            ),
             ..Default::default()
         };
 


hrkfdn:
Thanks a lot!

roderickvd:
This is now fixed in librespot-org/librespot@79f75b9119814254f854eeefe3b522517f80e80f.

hrkfdn:
Thanks @roderickvd, then we should revert this with the next librespot release.

PR #537, opened 2021-05-23
New Facebook Auth Flow

### LINKED ISSUE #310: Not able to login with Facebook
I didn't have any kind of trouble signing in with Facebook in the past.
I reinstalled today my Arch and cannot sign anymore, works ok with the Spotify GUI.
Any idea?

medwards:
@sashahilton00 and @michaelherger provided some guidance in https://github.com/librespot-org/librespot/discussions/635#discussioncomment-714466 on the new FB flow. This is a draft PR that attempts to implement OAuth2 Login with Facebook and would fix #310

Currently there's two problems:
* I probably need to request a client id and secret from Spotify (https://developer.spotify.com/documentation/general/), right now it fails getting the token because I'm using the desktop client id and not providing a secret.
* Once I get the token I'm not sure what to do with it. There's no immediately obvious way to turn it into `respot::Credentials`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -128,6 +128,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
 
 [[package]]
+name = "ascii"
+version = "1.0.0"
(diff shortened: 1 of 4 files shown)


medwards:
Closing to match #310. I've switched to using a "device password" until librespot adds support for the new auth flow (see librespot-org/librespot#754)

PR #540, opened 2021-05-28
Fix nerdfont pause/stop glyphs

rbuch:
b42315d accidentally changed both of these to the glyph for play, this reverts them back to the correct glyphs.


hrkfdn:
Oof, thanks!

Issue #23, opened 2019-03-07
Windows Support


Herbstein:
This is a tracking issue at this point.

I just tested after 41164df was commited, and #10 was closed. 

Cursive linking and compilation seems to work well now. 

The big issue is that right now librespot is configured to use pulseaudio as the backend. And, without fiddling with an MSYS2 installation, I don't think there's any way of making that work on Windows in a satisfactory way. However, it seems librespot want to move away from having their own internal playback implementation, and instead delegate it to the `Rodio` crate - discussed in librespot-org/librespot#195. This is being done in the librespot-org/librespot#277 pull request, though it seems progress has slowed a bit.

Looking at the `main.rs` file, I can also tell that you are making some assumptions about there being a `$HOME` environment variable, which you can't on Windows. Through a quick look around it seems [`app_dirs`](https://crates.io/crates/app_dirs) is a preferred way to find resource directories.

I didn't see anything else, but there might be some more lurking in there.

hrkfdn:
Some solid points, thanks. I'll keep it open so once I have some time I might be able to tackle some of these or maybe someone else would like to have a look.

Herbstein:
Just a small update. The Rodio backend for librespot almost kinda works. I'm experiencing some problems, but it seems the developer of the backend isn't. More testing is needed, but the prospects look great.

ashthespy:
For the impatient - the `portaudio` backend works quite fine with Windows - have it running here on Win10 a few months now now issues.

hrkfdn:
I have switched over to the Rodio backend, which is what librespot uses by default now. It should work on macOS and Windows out of the box. Reports are appreciated :)

hrkfdn:
I have moved over to PulseAudio by default again, as Rodio blocks a whole CPU thread in idle.

Optional linking with PortAudio/Rodio backend is still supported. I'm closing this for now.

terndrup:
Sorry to comment on a closed issue but how are you guys getting it to work on windows? My install or local build just fails with libdbus-sys errors

Herbstein:
I haven't been involved here for some time, but I got to the login screen just now with the following command:

```
cargo run --no-default-features --features rodio_backend,cursive/pancurses-backend
```

Make sure to use the MSVC-based rustc, and not the GNU-based one. Further, after logging in with Facebook I get a hang (looks like some infinite loop?), but after restarting it works perfectly and I get audio as expected.

terndrup:
[...]

PR #546, opened 2021-06-08
Add Windows instructions

### LINKED ISSUE #254: Windows support ?
since it requires 

libpulse-dev (or portaudio-dev, if you want to use the PortAudio backend)
libncurses-dev and libssl-dev
libdbus-1-dev
libxcb

Probably not.
But spotify + spicetify is eating my cpu :`(

rashil2000:
This follows from https://github.com/hrkfdn/ncspot/issues/23#issuecomment-603955270

Closes #254 

The CD can be setup for Windows releases too, see here https://github.com/42wim/ncspot/blob/master/.github/workflows/cd.yml

Once a release has been made, I'll make a PR in the scoop repos so that ncspot can be installed directly via scoop (like brew).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ strum_macros = "0.21.1"
 regex = "1"
 ioctl-rs = { version = "0.2", optional = true }
 serde_cbor = "0.11.1"
+pancurses = { version = "0.16.1", features = ["win32"] }
 
 [dependencies.cursive]
 version = "0.16.3"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -47,6 +47,7 @@ strum_macros = "0.21.1"
 regex = "1"
 ioctl-rs = { version = "0.2", optional = true }
 serde_cbor = "0.11.1"
+pancurses = { version = "0.16.1", features = ["win32"] }
```
hrkfdn, line 50: Will this break other platforms?

rashil2000: This `"win32"` feature flag affects only Windows builds, it will be ignored on other platforms.

hrkfdn:
Merged, thank you!

PR #563, opened 2021-07-11
add album cover to notification

charles-l:
First off, thanks for building such an awesome Spotify client. My poor desktop machine would chug with the official client but this one uses so much less memory and CPU time -- it's a world of difference.

This PR adds album cover images to the notification. (A small feature I've wanted :smiley: )

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -12,6 +12,9 @@ use crate::playable::Playable;
 use crate::spotify::Spotify;
 use crate::{config::Config, spotify::PlayerEvent};
 
+#[cfg(feature = "cover")]
+use crate::ui;
+
 #[derive(Display, Clone, Copy, PartialEq, Debug, Serialize, Deserialize)]
 pub enum RepeatSetting {
     #[serde(rename = "off")]
@@ -262,11 +265,14 @@ impl Queue {
             let mut current = self.current_track.write().unwrap();
             current.replace(index);
             self.spotify.update_track();
+
+            #[cfg(feature = "notify")]
             if self.cfg.values().notify.unwrap_or(false) {
-                #[cfg(feature = "notify")]
-                if let Err(e) = Notification::new().summary(&track.to_string()).show() {
-                    error!("error showing notification: {:?}", e);
-                }
+                std::thread::spawn({
+                    let track_name = track.to_string();
+                    let cover_url = track.cover_url();
+                    move || send_notification(&track_name, cover_url)
+                });
             }
         }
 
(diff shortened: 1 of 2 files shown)


charles-l:
Oh, just realized there was this function: https://github.com/hrkfdn/ncspot/blob/9eb99cc8be19f58299ef8eafdb0424c9ebc53b2e/src/ui/cover.rs#L294 -- I'll look into seeing if I can reuse it.

charles-l:
OK -- updated it to use the `ui::cover::download` function. And while I was at it, I had the cover code use the same cache path as the other cover code and pulled out the tempfile dependency.

src/queue.rs
```diff
@@ -403,3 +409,33 @@ impl Queue {
...
+        let path = ui::cover::cache_path_for_url(cover_url.to_string());
+
+        if !path.exists() {
+            if let Err(e) = ui::cover::download(cover_url.to_string(),
+            path.clone()) {
+                error!("Failed to download cover: {}", e);
+            }
+        }
+
+        Notification::new()
```
kepet19, line 426: I am not the developer of this project. 
I just toyed with the Idea aswell.
would it make sense to do the notification like this?
```rust
Notification::new()
            .summary(track_name)
            .body(&format!("{} - {}, album, artist))
            .icon(&path.to_str().unwrap())
            .show()
```
so the album and the artist shows up in the body of the notification.


It is actually pretty funny we both made almost the same code. (in the same code I mean both running in a new thread. your code is just so much cleaner!)

charles-l: Yeah, I think it's a good idea to include the artist as well -- fortunately, the Track::to_string function already does this: https://github.com/hrkfdn/ncspot/blob/aa8e9f10995bb71ecee78e2f7688deddbfbc0ebe/src/track.rs#L159-L163
It currently prints the artist followed by the track name in the notification. 

The current format (with this cover patch) will display a notification with the album art, artist name, and track name, exactly as the official Spotify client does. Including the album name by default might be too much. It's already a lot of text for some tracks. For example, in this example, it's already cut off at the edge of the screen with just the artist/track name.
image

I think it might be useful to be able to customize what the notification displays, but that goes beyond the scope of this PR -- maybe a follow up that adds a config for the notification format would be good...

kepet19: yeah that would make sense!

this version.
image


here is  how I have made it. Body notification shows artist and Summary show the track title.  

image

src/queue.rs
```diff
@@ -262,11 +265,14 @@ impl Queue {
...
+
+            #[cfg(feature = "notify")]
             if self.cfg.values().notify.unwrap_or(false) {
-                #[cfg(feature = "notify")]
-                if let Err(e) = Notification::new().summary(&track.to_string()).show() {
-                    error!("error showing notification: {:?}", e);
-                }
+                std::thread::spawn({
+                    let track_name = track.to_string().clone();
+                    let cover_url = track.cover_url().unwrap().clone();
```
hrkfdn, line 273: Is it guaranteed that Spotify will always provide a `cover_url`? If not, this `unwrap()` may panic.

hrkfdn (changes requested):
Thanks for the PR, generally it looks good! :+1: I have a minor change request regarding the `cover_url` unwrap. Also could you run `cargo fmt` to make sure the formatting is proper? Thanks a lot! :)

charles-l:
@hrkfdn sure! I've updated send_notification call to take an `Option<String>` instead of using `unwrap()`, and formatted my patch.

hrkfdn:
Merged, thanks!

PR #565, opened 2021-07-16
feat: Ability to set a custom config filename

mxdevmanuel:
Through `NCSPOT_CONFIG_FILE` env var a custom
configuration filename e.g "dark.toml" can
be specified

I use this with two different files "light.toml" and "dark.toml" to use different colors when my terminal colors are for daylight or nighttime

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,8 @@ Configuration is saved to `~/.config/ncspot/config.toml` (or `%AppData%\ncspot\c
 configuration during runtime use the `reload` statement in the command prompt
 `:reload`.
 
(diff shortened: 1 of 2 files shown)


hrkfdn:
Hey! Thanks for the PR, do you think a command line parameter would also be an option?

hrkfdn:
I have now added a flag for this `-c <filename>`, I hope this helps.

hrkfdn:
Hey! I'm closing this because this is now implemented as a command line flag. Thanks a lot for the idea + PR! :+1:

PR #576, opened 2021-08-16
Change volume when scrolling on it

thehabbos007:
Hey, really liking the project!

I missed being able to change volume by scrolling on the volume % display value.

I calculate the width of the volume display text and use the length of the string when checking for scroll mouse events.

Hope this is relevant :)
ncspot


hrkfdn:
Merged, thank you!

PR #598, opened 2021-09-09
Enable binding multiple commands to a key.

H-M-H:
This enables useful combinations of commands like `Space -> queue; move
down 1` using ';' as command separator. ';' can be escaped using ';;'.


hrkfdn:
Merged, thanks!

Issue #623, opened 2021-10-14
Get userRating in mpris metadata output


deed02392:
**Is your feature request related to a problem? Please describe.**
Currently, although we have the concept of saved/unsaved tracks, we don't expose that with the `userRating` metadata which is defined in the [latest mpris-spec](https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/).

**Describe the solution you'd like**
Map saved/unsaved to the maximum and minimum rating value according to the spec (float/double between 0.0 and 1.0).

**Additional context**
Following this, it would be nice to be able to update the saved/unsaved status of a track from the command line. This could be doable via dbus and would address feature request #524.

PR #624, opened 2021-10-14
Add `userRating` entity to mpris metadata

deed02392:
This PR enables MPRIS metadata output sent over D-Bus to output the saved/unsaved status of a track, with 0 and 1 corresponding to unsaved/not-liked and saved/liked in Spotify, respectively. This is in accordance with the [mpris-spec](https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/). Addresses #623.


deed02392:
Ran `cargo fmt` and amended the commit with the changes.

PR #627, opened 2021-10-15
fix inconsistency in help message and readme

Cardosaum:
Just a little fix in consistency. 
I don't know if there was a reason for `A` and `M` being the only ones without `Shift` prefixed.


hrkfdn:
Thanks, this is more explicit.

PR #644, opened 2021-11-13
Readme Overhaul

oclero:
I really like your tool, thanks for your beautiful work :)

Here is a PR that improves the README. I found it difficult to look for keybindings, so I made tables to make it easier to look for specific commands. You can have a preview of the resulting HTML-rendered README [here](https://github.com/oclero/ncspot/tree/readme-overhaul).

And a logo is included too!


hrkfdn:
Thanks a lot, it looks great and is a major improvement. The logo does too, though I'm afraid to use it due to it's similarity to the Spotify logo and these restrictions: https://developer.spotify.com/documentation/general/design-and-branding/#logos-and-naming-restrictions

I hope you understand. I appreciate you invested quite some time into this, but I'd rather not merge it with the logo to avoid potential problems. Sorry! :disappointed: Can we leave out the logo?

oclero:
Ha yes, I understand! Let me modify the logo a bit so we avoid any legal issue.

oclero:
By the way, they can't copyright an hex color. It makes no sense. Nothing prevents me to use their green, but not with their logo, of course.

oclero:
Do you prefer something like this?

ncspot-logo2-01

hrkfdn:
Looks good! Yeah true, I'm not even that worried about legal issues, but them blocking the client would be bad enough.

oclero:
Ok, I'll update the PR as soon as possible.

oclero:
@hrkfdn Done!

hrkfdn:
Merged! Thanks again.

Issue #633, opened 2021-10-29
Add Wayland clipboard support


mdartmann:
The clipboard package used unfortunately uses on X11's headers and library.  I mostly work on Wayland-based environments and XWayland's clipboard integration has been shotty at best for me, so I would appreciate it if it would be possible to support Wayland in a more native fashion, 

I don't much about rust myself (if I did, this would be a PR instead), but the [copypasta](https://crates.io/crates/copypasta) crate or the [wl-clipboard CLI tool](https://github.com/bugaevc/wl-clipboard) might be something to start working with.

Bettehem:
I found this project: [wl-clipboard-rs](https://github.com/YaLTeR/wl-clipboard-rs) which could be used to solve this issue. I could take a look at this and see if I can implement it (I'm still quite a Rust noob so not sure if I can), but it might be a couple weeks before I'll have the time to start

Bettehem:
I have now fixed this with my latest [pull request](https://github.com/hrkfdn/ncspot/pull/666).

PR #666, opened 2021-12-05
Added Wayland clipboard support.

Bettehem:
Hello, I added wayland clipboard support using the [wl_clipboard_rs](https://crates.io/crates/wl_clipboard_rs) crate.
Solves [#633](https://github.com/hrkfdn/ncspot/issues/633).

I tested with alacritty in both wayland and xorg and I also tested the primary clipboard and that seems to work as expected too.

### DIFF
--- a/src/sharing.rs
+++ b/src/sharing.rs
@@ -1,4 +1,16 @@
 #![cfg(feature = "share_clipboard")]
+use std::env;
+
+#[cfg(feature = "wayland_clipboard")]
+use {
+    std::io::Read,
+    wl_clipboard_rs::{
+        copy,
+        copy::{Options, Source},
+        paste,
+        paste::{get_contents, Error, Seat},
+    },
+};
 
 #[cfg(feature = "share_selection")]
 use clipboard::{x11_clipboard, x11_clipboard::X11ClipboardContext};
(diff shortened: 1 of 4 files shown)


src/sharing.rs
```diff
@@ -1,35 +1,159 @@
...
 #[cfg(not(feature = "share_selection"))]
 pub fn read_share() -> Option<String> {
-    ClipboardProvider::new()
-        .and_then(|mut ctx: ClipboardContext| ctx.get_contents())
-        .ok()
+    if std::option_env!("XDG_SESSION_TYPE").unwrap_or_default() == "wayland"
+        || !std::option_env!("WAYLAND_DISPLAY")
+            .unwrap_or_default()
+            .is_empty()
+        || std::option_env!("GDK_BACKEND").unwrap_or_default() == "wayland"
```
hrkfdn, line 25: Can this be extracted to a separate function for reuse/deduplication?

hrkfdn (changes requested):
Thanks a lot for your contribution :heart:, some minor changes are needed then we can get this in. I have one suggestion in the comments, the other is that this change should not break compilation on macOS and Windows. I haven't tested yet whether this is actually a problem, but I have updated the CI to run checks on both platforms.

src/sharing.rs
Bettehem: Thanks for your suggestion! I have now added an `is_wayland` function.

Bettehem:
> Thanks a lot for your contribution heart, some minor changes are needed then we can get this in. I have one suggestion in the comments, the other is that this change should not break compilation on macOS and Windows. I haven't tested yet whether this is actually a problem, but I have updated the CI to run checks on both platforms, so you can just rebase on `main` to test this.

I don't have any Windows or macOS machines to test on so I can't test this right now, but I could set up a Windows virtual machine to try this out

Bettehem:
I'm now setting up a windows vm to test building on it. I'll let you know how it goes asap

Bettehem:
I tested on windows now and the build is failing due to wl-clipboard-rs requiring a unix-like os to work. I'll try changing sharing.rs to only use the wl-clipboard-rs crate if a specific feature flag has been set.

Bettehem:
I have now fixed the issue. I also changed the wayland related environment variables to be checked at runtime instead of at compile time so you don't need to recompile when switching from Xorg to Wayland.\
I tested building on Windows with\
`cargo build --no-default-features --features rodio_backend,share_clipboard,cursive/pancurses-backend`\
and
`cargo build --no-default-features --features rodio_backend,share_clipboard,cursive/crossterm-backend`\
Both building and actually using the clipboard works.\
\
I also tested on Linux and if you build with the `wayland_clipboard` feature added, for example:\
`cargo build --features wayland_clipboard`\
\
Using `wayland_clipboard` will compile ncspot with the `wl-clipboard-rs` crate and use it if needed.

hrkfdn:
Merged, thank you!

PR #668, opened 2021-12-07
Added `jumpnext` and `jumpprevious` as commands

SlaynAndKorpil:
When I was trying to change my keybindings for ncspot in `config.toml`, I noticed there was no way to change the 'vim like' jump-to-next-search-result command from their default keybindings `"n"` and `"Shift+n"`.  
This was especially annoying since I wanted to remove the default keybindings with `default_keybindings = false`, to only have my own keybindings which would mean losing these two keybindings.

This commit adds the `"jumpnext"` and `"jumpprevious"` commands which generate a `Command::Jump(JumpMode::Next)` and `Command::Jump(JumpMode::Previous)` to mimic the functionality of the default `"n"` and `"Shift+n"` keybindings.

This also makes it possible to use these as commands in the commandline.


hrkfdn:
Thanks, merged! I added one small change to fix the serialization of jump commands as well :)

PR #682, opened 2021-12-26
Improved `seek` command to allow more units

cyqsimon:
I changed the `seek` command to use [parse_duration](https://docs.rs/parse_duration/latest/parse_duration/), so now it accepts not only milliseconds as input but also more convenient units (mix and match capability included). I think this is a good quality-of-life improvement.

new_seek

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -611,7 +611,7 @@ dependencies = [
  "lazy_static 1.4.0",
  "libc",
  "log",
- "num",
+ "num 0.3.1",
  "owning_ref",
  "syn",
  "unicode-segmentation",
@@ -1470,7 +1470,7 @@ dependencies = [
  "hyper-proxy",
  "librespot-protocol",
  "log",
- "num-bigint",
+ "num-bigint 0.4.3",
  "num-integer",
  "num-traits",
  "once_cell",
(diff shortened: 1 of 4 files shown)


hrkfdn (changes requested):
Thank you, love the idea! Could you fix the formatting using `cargo fmt`? Then I'll happily merge it.

cyqsimon:
Oh oops. I have a custom global `rustfmt.toml` so that screwed things up. Sorted.

Should I also commit an empty `rustfmt.toml` so that this doesn't happen again for others?

hrkfdn:
> Should I also commit an empty `rustfmt.toml` so that this doesn't happen again for others?

Ah, it only uses the default values if the project has an empty config? Thanks for the hint, I'll create one after merging this PR.

cyqsimon:
> > Should I also commit an empty `rustfmt.toml` so that this doesn't happen again for others?
> 
> Ah, it only uses the default values if the project has an empty config? Thanks for the hint, I'll create one after merging this PR.

I believe so. See https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#configuring-rustfmt

Says here that global configs are only sourced if no config file is found in project.

Issue #597, opened 2021-09-07
Unknown command: toggle & sort


cyqsimon:
## Description
`toggle` and `sort` are listed as available commands in [README](https://github.com/hrkfdn/ncspot#commands), but not actually usable.

## Steps to reproduce
1. Play any playlist
2. Type <kbd>:</kbd>, followed by either `toggle` or `sort`
3. UI prompts `ERROR: Unknown command: "toggle/sort"`

## System
 - OS: Manjaro KDE
 - Terminal: Yakuake & Konsole
 - Version: 0.8.2
 - Installed from: AUR & cargo

H-M-H:
Looks like `toggle` is actually `playpause` or one of its aliases: `pause`, `toggleplay`, `toggleplayback`.  Sorting a playlist works fine for me as long as I supply the sort key, for example `title`.

cyqsimon:
@H-M-H huh that's interesting. `sort` simply won't work on my box.

![bug](https://user-images.githubusercontent.com/28627918/132513076-5b97244c-a9ef-4056-bc61-3551b22687fe.gif)

H-M-H:
Do you have a playlist focused? As far as I can tell it only works for playlists.

cyqsimon:
I think it's fair to say I've tried to run `sort` in every view there is (including Queue [F1], Library-Tracks [F3, tab 1], Library-Playlist [F3, tab 4]). It hasn't worked on any view for me.

What exactly do you mean by "focused"? Can you perhaps record a GIF of the command working? Maybe I'm misunderstanding something.

H-M-H:
That's weird. With focused I mean you are viewing one of your saved playlists:
![2](https://user-images.githubusercontent.com/7461564/132537034-4dd5230a-b6f5-490b-8a11-a363bc1840d7.png)
![3](https://user-images.githubusercontent.com/7461564/132537048-e8ac6c0e-467b-4529-b574-951c8e98d044.png)

I am running on latest master though, so perhaps something changed in this regard?

cyqsimon:
Ah okay I see what I did wrong. I misunderstood you and was trying to sort the active queue. `sort` works as intended in the detailed view of the playlist.

Perhaps we can add a note in README? Or better yet, add the sort feature to the active queue as well?

cyqsimon:
[...]

PR #683, opened 2021-12-27
Improved README

cyqsimon:
- Regenerated table of contents using the Markdown All in One vscode plugin
- Fixed some spelling with cspell
- Documented some more commands

I also sneaked in an alias for `:repeat track` - `:repeat single`. It made sense to me while documenting the command so eh I just did it.


cyqsimon:
This addresses the documentation portion of #597.

I'll probably end up refactoring `command::parse(input: &str) -> Option<Vec<Command>>` into something like `command::parse(input:&str) -> Vec<Result<Command, CommandParseError>>`, which will allow more detailed feedback of what exactly was the problem with the input. So expect another PR from me soon =).

hrkfdn:
Merged, thank you! :)

PR #684, opened 2021-12-28
More detailed error message in case of command parse error

### LINKED ISSUE #597: Unknown command: toggle & sort

cyqsimon:
## Change Summary
- A failed command parse now returns more detailed error info, closes #597
- Added API:
  - `pub enum command::CommandParseError`
- Changed API:
  - `pub fn command::parse(&str) -> Option<Vec<Command>>` -> `pub fn command::parse(&str) -> Result<Vec<Command>, CommandParseError>`
- Breaking changes:
  - Some commands previously allow bad args silently and perform some default action. These cases will now fail fast.
    - Examples:
      - `shuffle onn` (misspelled `on`) no longer toggles shuffle.
      - `volup -10` (negative not allowed) no longer increases volume by 1%.
      - `sort title desending` (misspelled `descending`) no longer sorts in ascending.
    - Commands affected: `shuffle`, `repeat`, `save`, `volup`, `voldown`, `sort`
  - Continuous spaces in commands are now interpreted as one
    - Examples:
      - `sort title  descending` (two spaces between `args[0]` and `args[1]`, doesn't show up on rendered markdown) no longer sorts in ascending.
    - Potentially adverse consequences:
      - `newplaylist` can no longer create a playlist containing multiple continuous spaces.

## Feature Preview
feature-preview

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -309,8 +309,8 @@ pub fn parse(input: &str) -> Result<Vec<Command>, CommandParseError> {
             // https://docs.rs/regex/latest/regex/#example-avoid-compiling-the-same-regex-in-a-loop
             static ref CONTINUOUS_SPACE_MATCHER: Regex = Regex::new(r" +").unwrap();
         }
-        let commands_sanitised = CONTINUOUS_SPACE_MATCHER.replace_all(&command_input, " ");
-        let components: Vec<_> = commands_sanitised.trim().split(' ').collect();
+        let command_sanitised = CONTINUOUS_SPACE_MATCHER.replace_all(&command_input, " ");
+        let components: Vec<_> = command_sanitised.trim().split(' ').collect();
 
         let command = handle_aliases(components[0]);
         let args = components[1..].to_vec();
--- a/src/command.rs
+++ b/src/command.rs
@@ -99,12 +99,22 @@ impl fmt::Display for SeekDirection {
     }
 }
 
-#[derive(Display, Clone, Serialize, Deserialize, Debug)]
-#[strum(serialize_all = "lowercase")]
+#[derive(Clone, Serialize, Deserialize, Debug)]
 pub enum InsertSource {
-    Input(SpotifyUrl),
     #[cfg(feature = "share_clipboard")]
     Clipboard,
+    Input(SpotifyUrl),
+}
+
+impl fmt::Display for InsertSource {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let repr = match self {
+            #[cfg(feature = "share_clipboard")]
+            InsertSource::Clipboard => "".into(),
+            InsertSource::Input(url) => url.to_string(),
+        };
+        write!(f, "{}", repr)
+    }
 }
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
@@ -169,17 +179,14 @@ impl fmt::Display for Command {
             Command::Seek(direction) => format!("seek {}", direction),
             Command::VolumeUp(amount) => format!("volup {}", amount),
             Command::VolumeDown(amount) => format!("voldown {}", amount),
-            Command::Repeat(mode) => {
-                let param = match mode {
-                    Some(mode) => format!("{}", mode),
-                    None => "".to_string(),
-                };
-                format!("repeat {}", param)
-            }
-            Command::Shuffle(on) => {
-                let param = on.map(|x| if x { "on" } else { "off" });
-                format!("shuffle {}", param.unwrap_or(""))
-            }
+            Command::Repeat(mode) => match mode {
+                Some(mode) => format!("repeat {}", mode),
+                None => "repeat".to_string(),
+            },
+            Command::Shuffle(on) => match on {
+                Some(b) => format!("shuffle {}", if *b { "on" } else { "off" }),
+                None => "shuffle".to_string(),
+            },
             Command::Share(mode) => format!("share {}", mode),
             Command::Back => "back".to_string(),
             Command::Open(mode) => format!("open {}", mode),
@@ -205,7 +212,7 @@ impl fmt::Display for Command {
             },
             Command::Help => "help".to_string(),
             Command::ReloadConfig => "reload".to_string(),
-            Command::Insert(_) => "insert".to_string(),
+            Command::Insert(source) => format!("insert {}", source).trim().to_string(),
             Command::NewPlaylist(name) => format!("new playlist {}", name),
             Command::Sort(key, direction) => format!("sort {} {}", key, direction),
             Command::Logout => "logout".to_string(),
--- a/src/command.rs
+++ b/src/command.rs
@@ -305,12 +305,7 @@ pub fn parse(input: &str) -> Result<Vec<Command>, CommandParseError> {
 
     let mut commands = vec![];
     for command_input in command_inputs {
-        lazy_static! {
-            // https://docs.rs/regex/latest/regex/#example-avoid-compiling-the-same-regex-in-a-loop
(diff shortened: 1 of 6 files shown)


src/command.rs
```diff
@@ -271,226 +305,351 @@ pub fn parse(input: &str) -> Option<Vec<Command>> {
...
+                    Some('-') => i32::try_from(unsigned_millis)
+                        .map(|millis| SeekDirection::Relative(-millis)),
+                    _ => Ok(SeekDirection::Absolute(unsigned_millis)),
+                }
+                .map_err(|_| ArgParseError {
+                    arg: duration_raw,
+                    err: "Duration value too large".into(),
+                })?;
+                Command::Seek(seek_direction)
+            }
```
cyqsimon, line 530: @hrkfdn I would like to know your opinion on keeping backward compatibility for `seek [+|-]<milliseconds>`.  It caused me a lot of headache trying to write this code cleanly, but I still ended up with this mess.

If you think keeping backward compatibility here is not so important, I'm happy to go ahead and remove it.

src/command.rs
```diff
@@ -271,226 +305,351 @@ pub fn parse(input: &str) -> Option<Vec<Command>> {
...
+            "focus" => {
+                let &target = args.get(0).ok_or(InsufficientArgs {
+                    cmd: command.into(),
+                    hint: Some("queue|search|library".into()),
+                })?;
+                // TODO: this really should be strongly typed
+                Command::Focus(target.into())
```
cyqsimon, line 537: `Command::Focus` currently accepts a `String`. It is possible to strongly type it with an enum?

src/command.rs
```diff
@@ -271,226 +305,351 @@ pub fn parse(input: &str) -> Option<Vec<Command>> {
...
+            "reload" => Command::ReloadConfig,
             "insert" => {
-                if args.is_empty() {
-                    Some(Command::Insert(None))
-                } else {
-                    args.get(0)
-                        .map(|url| Command::Insert(Some((*url).to_string())))
-                }
+                // IDEA: this should fail fast too
+                Command::Insert(args.get(0).map(|&url| url.into()))
```
cyqsimon, line 575: Currently if the user inputs a bad URL, the command will run and fail silently without feedback. Ideally I would like to make it fail fast here too, but I know too little about the command executor to make edits confidently.

src/command.rs
```diff
@@ -271,226 +305,351 @@ pub fn parse(input: &str) -> Option<Vec<Command>> {
...
-            "noop" => Some(Command::Noop),
-            "redraw" => Some(Command::Redraw),
-            "exec" => Some(Command::Execute(args.join(" "))),
-            _ => None,
+            "noop" => Command::Noop,
+            "redraw" => Command::Redraw,
+            "exec" => Command::Execute(args.join(" ")),
+            _ => Err(NoSuchCommand {
+                cmd: command.into(),
+            })?, // I'm surprised this compiles lol
```
cyqsimon, line 650: I didn't think this would compile. Happy surprise.

src/main.rs
```diff
@@ -292,17 +292,19 @@ async fn main() -> Result<(), String> {
...
                     }
-                } else {
-                    let mut main = s.find_name::<ui::layout::Layout>("main").unwrap();
-                    let err_msg = format!("Failed to parse command(s): \"{}\"", cmd_without_prefix);
-                    main.set_result(Err(err_msg));
+                    Err(err) => {
+                        let mut main = s.find_name::<ui::layout::Layout>("main").unwrap();
+                        let err_msg = format!("Failed to parse command(s): {}", err);
+                        main.set_result(Err(err_msg));
+                    }
```
cyqsimon, line 307: This has the occasional problem where the error message's length exceeds the width of the terminal window, causing part of the message to be missing. Is there anything we can do about this?

cyqsimon:
Also I just took a look at my old issue and realised that this PR doesn't *really* close it in its current state. Commands like `sort`, if ran in an unsupported view, will still fail silently. This will require changes to the command executor.

src/main.rs
cyqsimon: I got rid of some wordiness, but the problem is not solved at the root. I'm thinking of some kind of auto-scroll or multi-line message.

src/command.rs
```diff
@@ -271,226 +305,351 @@ pub fn parse(input: &str) -> Option<Vec<Command>> {
 
     let mut commands = vec![];
     for command_input in command_inputs {
-        let components: Vec<_> = command_input.trim().split(' ').collect();
+        lazy_static! {
+            // https://docs.rs/regex/latest/regex/#example-avoid-compiling-the-same-regex-in-a-loop
+            static ref CONTINUOUS_SPACE_MATCHER: Regex = Regex::new(r" +").unwrap();
+        }
+        let command_sanitised = CONTINUOUS_SPACE_MATCHER.replace_all(&command_input, " ");
+        let components: Vec<_> = command_sanitised.trim().split(' ').collect();
```
hrkfdn, line 313: Maybe `str.split_whitespace()` is a simpler alternative we can use?

https://doc.rust-lang.org/std/primitive.str.html#method.split_whitespace

src/command.rs
hrkfdn: Hmm, I don't think it's so bad :smile: 

Ideally we should set the default unit to `ms` or fail when values without units are supplied, otherwise existing custom bindings will confuse users a lot, as they may suddenly seek forward by a couple of hundred seconds. So maybe we should keep it backward compatible for now? WDYT?

src/command.rs
hrkfdn: I have no preference here, both are fine to me. I think the gain is rather low compared to the effort, as `Layout` already checks if the screen exists when processing the command, but feel free to type it if you prefer.

src/command.rs
hrkfdn: Ideally yes, we could at least validate the URL or check if it can be parsed using `SpotifyUrl::from_url`. Everything after that requires API calls.

hrkfdn (commented):
Thanks for tackling this. I have added some comments.

src/command.rs
cyqsimon: The reason I suggested to make `second` the default unit is actually quite simple... `parse_duration` crate doesn`t provide any options to configure its default unit, which means we either have to use its default (second), or do the dance we've got here to try parsing to `ms` first.

If you think the code isn't too terrible I'm happy to keep it as is.

src/command.rs
cyqsimon: Oh bollocks. I knew about this function for some time now, but never understood its point nor have read its documentation. It is better indeed. Thanks for showing me the way.

src/command.rs
cyqsimon: You're right. It's probably more effort than it's worth unless we're actually considering adding another screen. This can wait for when (and if) that situation arises.

src/main.rs
cyqsimon: I took another look today after simply shortening the message and, eh, it's good enough. Too much work than it's worth for now.

cyqsimon:
Okay. I'm pretty happy with the state this PR is in. Problems in #597 are more or less sufficiently addressed right now.

I've also done a bit of code cleanup (e.g. sorting match branches), so hopefully this will make the codebase slightly easier to maintain.

hrkfdn:
Thanks again, looks very good to me. Let's get this in!

PR #688, opened 2022-01-01
Remove panics in the update_token flow.

varrocs:
Hello
This pull request fixes a crash:
1. Start ncspot
2. Put the computer to sleep for an hour
3. Try to search for something

In this case the app crashed. The root cause of the crash is that the `response` in `get_token` holds a `MercuryError` instead of a result and the `unwrap` panics

With this fix the errors in the response processing are handled gracefully and the app doesn't crash.

The first search still fails after the sleep but the app doesn't crash

### DIFF
--- a/src/spotify_api.rs
+++ b/src/spotify_api.rs
@@ -77,19 +77,23 @@ impl WebApi {
             .as_ref()
         {
             channel.send(cmd).expect("can't send message to worker");
-            let token = futures::executor::block_on(token_rx).unwrap();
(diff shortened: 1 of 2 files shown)


varrocs:
I think this also fixes this issue: https://github.com/hrkfdn/ncspot/issues/552

hrkfdn:
Looks like the token renewal coordination should be improved, but we should definitely handle this scenario gracefully. Thanks for the PR, merged!

PR #695, opened 2022-01-04
`seek` spits out correct error when used with no args

cyqsimon:
Just a minor fix on my code in #684


hrkfdn:
Merged, thank you!

PR #706, opened 2022-01-13
Remove notifications' dependency on 'cover'

rashil2000:
This PR, in essence, just moves the functions `download` and `cache_path_for_url` from cover.rs to utils.rs. This is done so that the notification icon can be used on Windows. These functions were not doing doing anything Unix-specific, but since the `cover` feature is disabled on Windows, they couldn't be used before.

I have also enabled notifications and clipboard features for CI/CD on Windows. I have tested them in all scenarios, but please let me know if I missed anything.


hrkfdn:
Merged, thank you!

PR #709, opened 2022-01-14
ui: listview: fix log::info errors

isinyaaa:
This commit fixes the errors below, introduced in "Move from
`ScrollBase` to `scroll` module" [1]. As noted by the rust docs [2],
named arguments should be used in the format

```rust
log::info!("content len: {content_len}", content_len=content_len);
```

As this provides no advantage over unnamed args, use them for
conciseness with the rest of the project.

```raw
error: there is no argument named `content_len`
  --> src/ui/listview.rs:88:34
   |
88 |         log::info!("content len: {content_len}");
   |                                  ^^^^^^^^^^^^^

error: there is no argument named `loaded`
   --> src/ui/listview.rs:100:35
    |
100 |         log::info!("can paginate: {loaded}");
    |                                   ^^^^^^^^

error: could not compile `ncspot` due to 2 previous errors
```

[1] - Move from `ScrollBase` to `scroll` module (adf8da2f)
[2] - https://doc.rust-lang.org/std/fmt/#named-parameters

Signed-off-by: Isabella Basso <isabbasso@riseup.net>

### DIFF
--- a/src/ui/listview.rs
+++ b/src/ui/listview.rs
@@ -85,7 +85,7 @@ impl<I: ListItem> ListView<I> {
 
     pub fn content_height_with_paginator(&self) -> usize {
         let content_len = self.content.read().unwrap().len();
-        log::info!("content len: {content_len}");
+        log::info!("content len: {}", content_len);
 
         // add 1 more row for paginator if we can paginate
         if self.can_paginate() {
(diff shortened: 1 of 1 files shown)


isinyaaa:
I'm sorry, I hadn't yet updated to rustc 1.58.0 [1] and proceeded to check (slightly) outdated docs.

[1] - https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html

Closing...

PR #718, opened 2022-01-21
Use notify-rust with dbus instead of zbus

jirutka:
This project already uses "dbus" crate; bundling two dbus API implementations in the ncspot binary doesn't make much sense, it just increases the binary size.


hrkfdn:
Makes sense! Merged, thank you.

PR #721, opened 2022-01-23
Add a link to the full command list

ChaboCode:
Added a link to /src/command.rs highlighting lines 227 to 265 which have the commands. Sugested by @MrConorAE


MrConorAE:
Yep, exactly! Thanks.

hrkfdn:
Thanks for the PR, but I'd prefer to not merge this. It's very technical and the link would need an update every time the line/position changed, resulting in outdated documentation.

PR #741, opened 2022-02-08
Improve playlist sorting

Roger-Roger-debug:
Add secondary and tertiary sorting options to behave more like Spotify when sorting by artist or album.


hrkfdn:
Thanks, that's a nice improvement!

PR #780, opened 2022-04-07
Only download cover image when flag enabled

rbuch:
Previously, cover images were downloaded and cached even when the corresponding feature flag was disabled; this adds a check to make sure that the cover image is actually used before trying to get and store it.


hrkfdn:
Merged, thanks!

Issue #786, opened 2022-04-18
Context Menu for track using spotify link in clipboard does not have the option to play the song


maximekabel:
**Is your feature request related to a problem? Please describe.**
When using `Ctrl+V` to open the context menu of a track who's spotify link is in the clipboard, there is no option to play the track.

**Describe the solution you'd like**
A solution for this could be to add `Play track` option in the context menu.

Bettehem:
I'm working on this now. If no problems arise, it could be done by the end of the day.

PR #790, opened 2022-04-20
Added option to play track from context menu dialog

Bettehem:
Hello again!
I have added an option to the context menu dialog which allows to play the track or add to queue. This solves [#786](https://github.com/hrkfdn/ncspot/issues/786).

There's one problem currently though. In the dialog where you can select to play now or add to queue, only arrow keys work.
When trying to use j and k, it scrolls the list in the background and not in the dialog, so this probably shouldn't be merged until that is solved.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -285,17 +285,26 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "3.1.8"
+version = "3.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "71c47df61d9e16dc010b55dba1952a57d8c215dbb533fd13cdd13369aac73b1c"
+checksum = "6aad2534fad53df1cc12519c5cda696dd3e20e6118a027e24054aea14a0bdcbe"
 dependencies = [
  "atty",
  "bitflags",
+ "clap_lex",
  "indexmap",
- "os_str_bytes",
  "strsim",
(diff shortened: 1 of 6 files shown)


hrkfdn:
Hi there, thanks for the PR. Would love to merge this, but it appears to contain commit https://github.com/hrkfdn/ncspot/pull/790/commits/9a97a6ebdf11b0519dd0275fd5df4f4bd6055a22 from your other PR. Could you remove that commit?

Bettehem:
Oh, my bad, I hadn't noticed that. I have removed it now.

hrkfdn:
Thank you! Merged :)

Bettehem:
Nice! There's a problem though, `j` and `k` don't work in the new dialog, only arrow keys do. `j` and `k` scrolls the list in the background of the dialog and not inside the dialog itself. Any suggestions how to fix this?

hrkfdn:
@Bettehem: Ah, should be fixed with https://github.com/hrkfdn/ncspot/commit/52e2a5383945fb4e1b6a3286570a2d9e13f55c3f :)

Two things were missing: the type in `handle_move_command` for the select items was not matching and commands were not forwarded to the context menu in the command handler. There's still a lot of boilerplate involved when creating context menus :(

Bettehem:
Oh I see! This is good to know if I work with context menus in the future. Thanks!

Issue #730, opened 2022-01-26
Configuration option for changing the name order of a song


haukkagu:
Ncspot displays the name of the artist before the name of the track for every song (e.g. "Ludwig van Beethoven - Für Elise"). This is normally just a minor annoyance for me, but when listening to albums of many authors everything becomes a cluttered mess ([example](https://i.imgur.com/TUVcbUE.png)). I prefer that the name of the track comes first instead of the name of the artist (e.g "Für Elise - Ludwig van Beethoven" instead of "Ludwig van Beethoven - Für Elise"). This is basically the norm on every other music client, including the official spotify client, so I'd assume that many other people prefer it like that as well. I think there should atleast be a configuration option for changing this (maybe an option called "track_name_first" that can be set to true or false?).

Thanks!

hrkfdn:
Solved via #800

Issue #558, opened 2021-06-24
`ListView` customization


emekoi:
would it be possible to customize the list view for the queue and/or the library so that we could omit or add relevant columns? for example, i am almost never interested in the album field, but it's always there taking up space. perhaps there could be a config option that holds the wanted columns in an array.

hrkfdn:
Solved via #800

PR #800, opened 2022-05-01
Add `track_format` config option

Bettehem:
Hello again!
I added a new config option: `tracklist_formatting`.
This solves #730 and #558


hrkfdn:
Thanks for the PR, appreciate it. I think a more flexible/customizable approach would be the one described here: https://github.com/hrkfdn/ncspot/issues/558

It would allow custom formats for tracks (and maybe albums and other types) so users can decide for themselves, but is probably a little bit more complex.

hrkfdn:
Thanks for the PR, appreciate it. I think a more flexible/customizable approach would be the one described here: https://github.com/hrkfdn/ncspot/issues/558

It would allow custom formats for tracks (and maybe albums and other types) so users can decide for themselves, but is probably a little bit more complex.

Bettehem:
Oh yeah, that sounds like a good idea! I could try changing it to work like that🤔

Bettehem:
Hello, I have made some some modifications now. I added an `active_fields` config option which lets you configure what should be shown and in which column. This also removes the need for a separate `album_column` option. I also added some examples to the README. This should now solve both #730 and #558

src/model/track.rs
```diff
@@ -159,19 +161,105 @@ impl From<&SavedTrack> for Track {
...
+                    config::ActiveField::Title => title_index = i,
+                    config::ActiveField::Artists => artists_index = i,
+                    _ => {}
+                }
+            }
+            if title_index < artists_index {
+                write!(f, "{} - {}", self.title, self.artists.join(", "))
+            } else {
+                write!(f, "{} - {}", self.artists.join(", "), self.title)
+            }
```
hrkfdn, line 201: There is a lot of logic here to generate the output string for the columns. I'm wondering if there is a more dynamic approach, i.e. the user can provide something like `format_left = "%artists - %title"` which is then replaced.

hrkfdn (commented):
Thanks for reiterating and tackling this. I have a suggestion in the comment below, it would also apply to the other columns. What do you think?

src/model/track.rs
Bettehem: Ok, I like the idea. It would probably make the logic simpler. I'll see what I can do!

Bettehem:
Hi, Sorry for taking so long, I have been busy with other stuff, but I have now made a more flexible system for customizing the formatting. What do you think?

hrkfdn:
Thanks, looks pretty good already. I've rebased the latest `main` and did a small refactoring. I'm wondering if we should support formatting for podcast episodes as well? I think they may stick out if you have a playlist with a mix of tracks and podcasts..

Bettehem:
Oh, true. I didn't know you could mix podcasts and songs in the same playlist 🤦
But yeah the formatting should probably work for podcast episodes too if they can be added to the same list.
Thanks for the feedback again, I'll try to make it work for episodes too!

Bettehem:
Hmm, I was just testing, and noticed that the custom formatting already works for Episodes in playlists, because episodes in playlists are being treated as tracks. I made a new commit which moves the custom format function to Playable impl so if in the future playlists differentiate between Episodes and Playlists its easier to implement

hrkfdn:
> Hmm, I was just testing, and noticed that the custom formatting already works for Episodes in playlists, because episodes in playlists are being treated as tracks.

Huh, indeed. Sounds like a bug, I've just checked and it looks like either rspotify or the Spotify API returns them all as tracks. I have reported this upstream at rspotify: https://github.com/ramsayleung/rspotify/issues/320

> I made a new commit which moves the custom format function to Playable impl so if in the future playlists differentiate between Episodes and Playlists its easier to implement.

That makes sense, thanks.

I think we can merge this now. I was wondering whether we should rename the config variables to `track_formatting` as it's a little bit shorter and reflects the affected playable type better. What do you think?

Once the bug is fixed we can also support custom podcast formatting, which may have different fields. Also in the future we could add support custom formatting for other playable types, i.e. albums.

hrkfdn:
I have changed the naming and merged it. Thanks for your contribution and considering all the annoying comments :smile:

Bettehem:
> I have changed the naming and merged it. Thanks for your contribution and considering all the annoying comments 😄

Nice, and thank you for taking the time to give feedback and suggestions, I'm still a Rust noob so it has been very helpful!

PR #821, opened 2022-06-03
Fix some errors and warnings when enabling/disabling features

cyqsimon:
Most were just unused imports and lint warnings, but `sharing.rs` was erroring with `wayland_clipboard` feature enabled.

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -138,6 +138,7 @@ pub enum Command {
     VolumeDown(u16),
     Repeat(Option<RepeatSetting>),
     Shuffle(Option<bool>),
+    #[cfg(feature = "share_clipboard")]
     Share(TargetMode),
     Back,
     Open(TargetMode),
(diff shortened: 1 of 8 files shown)


cyqsimon:
I checked each feature individually and fixed what I found. But it's just too much work to exhaustively check all possible combinations of features.

I'm thinking there must be a better way and/or existing tool to exhaustively test all combinations right? It sounds like a very common demand that many large projects will face. If anyone knows of such a tool please let me know. Thanks.

cyqsimon:
Rebased on `main`.

hrkfdn:
Merged, thank you for the PR.

You're right, it is quite cumbersome. Generally speaking the default features are the most critical. Some of the optional features were submissions to satisfy individual needs and are not really supported by me (i.e. the `cover` feature). But you're right. some more CI checks wouldn't hurt.

PR #822, opened 2022-06-03
Add desktop entry & notification icon hint

cyqsimon:
- Added a `.desktop` file so that the user can launch `ncspot` straight from the launcher of their desktop environment
- Added the desktop entry hint to notification so ncspot's logo can be displayed inline with the notification

On KDE Plasma:
- Before
before
- After
after

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -88,5 +88,6 @@ It is heavily inspired by ncurses MPD clients, such as ncmpc."""
 license-file = ["LICENSE"]
 assets = [
     ["target/release/ncspot", "usr/bin/", "755"],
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -88,5 +88,6 @@ It is heavily inspired by ncurses MPD clients, such as ncmpc."""
 license-file = ["LICENSE"]
 assets = [
     ["target/release/ncspot", "usr/bin/", "755"],
+    ["misc/ncspot.desktop", "usr/share/applications/", "644"],
```
cyqsimon, line 91: I'm not very familiar with the debian packaging manifest, so I'm not sure if this is good enough.

hrkfdn:
Merged, thank you!

PR #855, opened 2022-07-08
Add 'hide_display_names' config option

### LINKED ISSUE #853: Option to hide own username
I infrequently share my full screen under a pseudonym and would like 'Library of Real Name' and 'Playlist Name * Real Name' to not be displayed in ncspot. Some users with Spotify linked to their Facebook account are currently not given the option to unlink their accounts and as such their display name will always be their real name (or at least until the unlink option is added for them).

As the name being displayed was previously implemented as a feature on #652 , perhaps this could be made an option in the config?  

The following are "work-arounds" but they don't really address the issue.
1. Just don't have ncspot open while sharing the screen. This is simple enough, but doesn't really address the issue at hand.
2. Move to a view which doesn't have the user's name displayed such as "Search" or a playlist made by another user. This provides some limited functionality of the application but is better than the previous option.
3. Migrate to a Spotify account which is not connected to Facebook. I've been considering doing this, but figured I should ask about this feature request before going through the process of transferring everything to a new account.

sudo-jaa:
Adds an option to the `config.toml` that allows a user to disable the display of usernames in the main library tab and against playlists.

This also updates the README.md file to account for this additional config option. 

This implementation follows a similar pattern to configuring custom [library_tabs](https://github.com/hrkfdn/ncspot/blob/main/src/ui/library.rs#L25), and custom [track_format](https://github.com/hrkfdn/ncspot/blob/main/src/model/track.rs#L186)

Fixes #853

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,31 +325,32 @@ runtime use the `reload` command.
 
 Possible configuration values are:
 
-| Name                            | Description                                       | Possible values                                                 | Default     |
-|---------------------------------|---------------------------------------------------|-----------------------------------------------------------------|-------------|
-| `command_key`                   | Key to open command line                          | Single character                                                | `:`         |
-| `initial_screen`                | Screen to show after startup                      | `"library"`, `"search"`, `"queue"`, `"cover"`<sup>[1]</sup>     | `"library"` |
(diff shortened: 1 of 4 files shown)


hrkfdn:
Merged, thank you! :)

PR #868, opened 2022-07-20
Added possibility to Follow/Unfollow artists via the context menu

Bettehem:
Hello, I noticed that there was no easy way of following/unfollowing artists unless their tracks were in your library already, so I added a new option to the context menu. Now when pressing on "Show artist", you will have the option to select if you want to view the artist page or if you want to follow/unfollow the artist.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -11,7 +11,9 @@ use crate::library::Library;
 use crate::queue::{Queue, RepeatSetting};
 use crate::spotify::{Spotify, VOLUME_PERCENT};
 use crate::traits::{IntoBoxedViewExt, ViewExt};
-use crate::ui::contextmenu::{AddToPlaylistMenu, ContextMenu, PlayTrackMenu, SelectArtistMenu};
+use crate::ui::contextmenu::{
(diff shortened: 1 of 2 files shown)


hrkfdn:
Hey, thanks again. I'm not sure how intuitive it is to find this action after pressing "Show artist". I feel like this could simply be an extra item in the root section of the context menu, i.e. as "Follow artist" below "Save track". Would you be fine with this as well?

Bettehem:
Sure, I can change that!

Bettehem:
Hmm, I think it might be better behind the "Show artist" button after all. Perhaps the "Show artist" text could be changed to "Artist", or "Artists" if there are many. The problem with having the "Follow artist" text in the root section of the context menu is that if a track has multiple artists, the correct text could be "Follow" for one artist, but should be "Unfollow" for another.

hrkfdn:
It could be "(Un)follow artists" with a submenu allowing you to follow or unfollow the track artists, depending on the current state. But I'm also fine with your idea, though I think we should rename it from "Show artist" to something else.

Bettehem:
Ok, Hello again!
I have now changed "Show Artist" to "Artist" or "Artists" if there are many. What do you think?

hrkfdn:
Thanks a lot 🚀! Sorry for the delay, let's get it in!

Issue #830, opened 2022-06-09
Wayland: Nothing copied into the clipboard on track share


cartesius68:
Pressed `X` or `Shift-X` to copy the track URL and nothing has been copied into the clipboard.

* Arch Linux
* SwayWM 1.7 or GNOME 42 (on **Wayland**)
* ncspot `commit ce6b360cba90967`
* built via `cargo build`

hrkfdn:
Did you build with the `wayland_clipboard` feature enabled?

jghauser:
I just realised that this is the reason why copying to clipboard didn't work for me. Would you consider enabling this option by default? Many if not most Linux systems will be on wayland nowadays.

Bettehem:
Yeah, the `wayland_clipboard` would probably be good to be enabled by default, but iirc it wouldn't compile on Windows with that feature enabled. I'm not that experienced with Rust or Cargo, but perhaps it's possible to exclude the feature on platforms other than Linux?

hrkfdn:
Fixed with #869

PR #869, opened 2022-07-20
Enable wayland clipboard support by default on supported platforms

Bettehem:
Hi, I attempted to solve #830. This should now automatically build the wayland_clipboard clipboard feature on supported platforms.
\
\
\
I noticed right after pushing this branch that I included commits from my other PR here too, but I reverted those, so it should be fine now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ tokio-stream = "0.1.9"
 toml = "0.5"
 unicode-width = "0.1.9"
 url = "2.2"
+[target.'cfg(all(target_family = "unix", not(any(target_os = "macos", target_os = "ios"))))'.dependencies]
 wl-clipboard-rs = {version = "0.6", optional = true}
 
 [dependencies.rspotify]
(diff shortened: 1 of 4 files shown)


hrkfdn:
Hey, thanks for the PR. Generally I'm not against enabling Wayland clipboard support by default, but I'll have to check how this affects the *BSDs, as they would fall under `unix` but don't use Wayland.

Bettehem:
Wayland should work on the BSDs too:
https://docs.freebsd.org/en/books/handbook/x11/#x-wayland
https://en.wikipedia.org/wiki/Wayland_(display_server_protocol)

hrkfdn:
> Wayland should work on the BSDs too: https://docs.freebsd.org/en/books/handbook/x11/#x-wayland https://en.wikipedia.org/wiki/Wayland_(display_server_protocol)

Unfortunately for OpenBSD and NetBSD that is not the case. I also don't know how many FreeBSD users use Wayland these days, as I believe it's still somewhat experimental. But I haven't used FreeBSD in a while.

Bettehem:
I found this article of wayland running on NetBSD too so apparently it can work: https://blog.netbsd.org/tnf/entry/wayland_on_netbsd_trials_and

But I have now made changes so that the wayland clipboard feature is only built by default on linux, as the wayland support in the BSDs doesn't seem to be official.

hrkfdn:
Thanks! Let's get it in! :)

hrkfdn:
I just wanted to share a link on GNOME and it seems the Wayland clipboard protocol implementation used by the `wl-clipboard-rs` crate [does not support GNOME](https://github.com/YaLTeR/wl-clipboard-rs/issues/8) :( 

Is there another crate that we can use for this will support GNOME and the wlroots-based compositors?

Bettehem:
Oh, interesting... I looked into this, and it seems to be the best course of action to default to the x11 clipboard if gnome is detected. Then it works as expected. I made a pull request which fixes the issues on gnome: #912

PR #894, opened 2022-08-05
Switch Cursive backend defualt to termion for true color support

eulerfan271:
Cursive supports true color support (rather than just the 256 terminal colors), but not with all backends. ncspot's default does not appear to be one that supports true color.

ncspot already has a feature flag for termion, which does support true color, so I suggest making that Cursive backend the default.


hrkfdn:
Thanks, let's give it a try.

PR #912, opened 2022-08-17
Fixed clipboard not working on GNOME Wayland.

Bettehem:


hrkfdn:
Thanks for taking care :pray:, not sure about KDE, but this fixes it for GNOME at least!

PR #916, opened 2022-08-24
Fix back button mouse left click

### LINKED ISSUE #915: [Mouse Support] Mouse press on back button doesn't go back one layer
**Describe the bug**
The mouse button doesn't behave like the backspace on the keyboard. When there are two layers and the user is in the second layer, pressing the back button will go back to the library view, not the previous view in the tab the user is currently in. The only example of this that I can find right now happens in Browse, because that's the only place that goes two layers deep (Library -> Genre, Genre -> Songs). 

**To Reproduce**
1. Go to 'Browse'
2. Click any Genre
3. Click any item in the genre
4. Try to go back with the mouse

**Expected behavior**
Expected: Go up one layer
Actual: Go back to the library

 - OS: Arch Linux
 - Terminal: Alacritty
 - Version: 0.11.0
 - Installed from: AUR

**Additional context**
Love the new mouse support :heart:

ThomasFrans:
fixes #915

### DIFF
--- a/src/ui/layout.rs
+++ b/src/ui/layout.rs
@@ -3,7 +3,7 @@ use std::time::{Duration, SystemTime};
 
 use cursive::align::HAlign;
 use cursive::direction::Direction;
-use cursive::event::{AnyCb, Event, EventResult, MouseButton};
+use cursive::event::{AnyCb, Event, EventResult, MouseButton, MouseEvent};
 use cursive::theme::{ColorStyle, ColorType, Theme};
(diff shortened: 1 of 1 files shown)


hrkfdn:
Nice catch, thanks. I'm guessing this is because it's triggered twice, on mouse press and release. `button()` will return a result in both scenarios: https://docs.rs/cursive_core/0.3.4/src/cursive_core/event.rs.html#459

PR #919, opened 2022-08-25
Fix order of parameters to Cache::new

rbuch:
When updating to librespot 0.4.0 in c41294c, the volume and audio path
parameters were specified in the wrong order.

This broke the `audio_cache = false` case, as well as caching files in the `volume/` folder rather than the `files/` folder.


hrkfdn:
Good catch, thank you!

Issue #131, opened 2020-02-23
1 more items blinks...


KarlZeilhofer:
Hi, 

i just searched for `thomas bergersen` and the results show a blinking `1 more items, scroll to load` under Artists. 

![2020-02-23_001](https://user-images.githubusercontent.com/2773445/75119965-d0b02580-5687-11ea-8e9d-8b54fd6dce0d.png)

Seems to be a bug. 

Using v0.1.2 on Linux Mint 19.2

Greets, Karl

hrkfdn:
Well.. this one is odd. The web API actually returns a result count of 3, but only returns 2 items. Seems like a bug on Spotify's end.. :thinking: 

```json
{
  "artists": {
    "href": "https://api.spotify.com/v1/search?query=thomas+bergersen&type=artist&offset=0&limit=20",
    "items": [
      {
        "external_urls": {
          "spotify": "https://open.spotify.com/artist/6BF0bXbsdujMSMeFZBGcBq"
        },
        "followers": {
          "href": null,
          "total": 145260
        },
        "genres": [
          "epicore",
          "soundtrack",
          "video game music"
        ],
        "href": "https://api.spotify.com/v1/artists/6BF0bXbsdujMSMeFZBGcBq",
        "id": "6BF0bXbsdujMSMeFZBGcBq",
        "images": [
          {
            "height": 1000,
            "url": "https://i.scdn.co/image/821040496f31cebb75daec6bedc2c5fd9020b259",
            "width": 1000
          },
          {
            "height": 640,
            "url": "https://i.scdn.co/image/9cc2b715294ec21bebc2629a257d302b98ec9ba6",
            "width": 640
          },
          {
            "height": 200,
            "url": "https://i.scdn.co/image/57215347e85ecb6f0b139a2d6f68254fbc815ee9",
            "width": 200
          },
          {
            "height": 64,
            "url": "https://i.scdn.co/image/ed3103d6db68f4c95b96ccf8f4519546aa1c1739",
            "width": 64
          }
        ],
        "name": "Thomas Bergersen",
        "popularity": 71,
        "type": "artist",
        "uri": "spotify:artist:6BF0bXbsdujMSMeFZBGcBq"
      },
      {
        "external_urls": {
          "spotify": "https://open.spotify.com/artist/5zkXfTS1MSmo2yPhCLh603"
        },
        "followers": {
          "href": null,
          "total": 116
        },
        "genres": [],
[...]

PR #939, opened 2022-09-14
Fixes a pagination bug that caused items to not load when scrolling with the mouse

### LINKED ISSUE #938: Scroll to load extra items no longer works with the mouse
**Describe the bug**
When there are a lot of items in a listview, there is a text message that says to scroll to get more items loaded. It doesn't work anymore in the most recent version. I think the #918 fix caused this as a side effect.

**To Reproduce**
1. Go to a long podcast (I used "De Universiteit van Vlaanderen Podcast", there's probably normal podcasts as well but this is just the one that triggered the bug for me :smile: )
2. Scroll to the bottom with the mouse

**Expected behavior**
The extra items load at the bottom.

**Screenshots**
image

 - OS: Arch Linux
 - Terminal: Alacritty
 - Version: Latest git version
 - Installed from: AUR

ThomasFrans:
This PR fixes the pagination but that prevented loading. I also added some documentation to some of the less obvious methods. There is still a bug present that causes the pagination message to appear at the bottom of the Browse tab, even when there isn't any more items. Help would be wanted with fixing that. I could also open another separate bug report for that if that's easier :slightly_smiling_face: .

fixes #938

### DIFF
--- a/src/ui/listview.rs
+++ b/src/ui/listview.rs
@@ -59,7 +59,7 @@ impl<I: ListItem> Scroller for ListView<I> {
 
 impl<I: ListItem> ListView<I> {
     pub fn new(content: Arc<RwLock<Vec<I>>>, queue: Arc<Queue>, library: Arc<Library>) -> Self {
-        Self {
+        let result = Self {
             content,
             last_content_len: 0,
             selected: 0,
@@ -72,7 +72,9 @@ impl<I: ListItem> ListView<I> {
             library,
             pagination: Pagination::default(),
             title: "".to_string(),
-        }
+        };
+        result.try_paginate();
+        result
     }
 
     pub fn with_title(mut self, title: &str) -> Self {
@@ -99,8 +101,28 @@ impl<I: ListItem> ListView<I> {
         }
     }
 
+    /// Return wether there are still items that aren't shown in the listview.
+    ///
+    /// `true` if there are unloaded items
+    /// `false` if all items are loaded
     fn can_paginate(&self) -> bool {
         self.get_pagination().max_content().unwrap_or(0) > self.get_pagination().loaded_content()
+    }
+
+    /// Try to load more items into the list if neccessary.
+    #[inline]
+    fn try_paginate(&self) {
+        // Paginate if there are more items
+        //  AND
+        //   The selected item is the current last item (keyboard scrolling)
+        //    OR
+        //   The scroller can't scroll further down (mouse scrolling)
+        if self.can_paginate()
+            && (self.selected == self.content.read().unwrap().len().saturating_sub(1)
+                || !self.scroller.can_scroll_down())
+        {
+            self.pagination.call(&self.content, self.library.clone());
+        }
     }
 
     pub fn get_selected_index(&self) -> usize {
@@ -297,7 +319,10 @@ impl<I: ListItem> View for ListView<I> {
             Event::Mouse {
                 event: MouseEvent::WheelDown,
                 ..
-            } => self.scroller.scroll_down(3),
+            } => {
+                self.scroller.scroll_down(3);
+                self.try_paginate();
+            }
             Event::Mouse {
                 event: MouseEvent::Press(MouseButton::Left),
                 position,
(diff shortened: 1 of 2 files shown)


src/ui/listview.rs
```diff
@@ -72,7 +72,9 @@ impl<I: ListItem> ListView<I> {
             library,
             pagination: Pagination::default(),
             title: "".to_string(),
-        }
+        };
+        result.try_paginate();
```
hrkfdn, line 76: Hmm, could you explain why this change is necessary?

src/ui/listview.rs
```diff
@@ -99,10 +101,30 @@ impl<I: ListItem> ListView<I> {
...
+    #[inline]
+    fn try_paginate(&self) {
+        // Paginate if there are more items
+        //  AND
+        //   The selected item is the current last item (keyboard scrolling)
+        //    OR
+        //   The scroller can't scroll further down (mouse scrolling)
+        if self.can_paginate()
+            && (self.selected == self.content.read().unwrap().len().saturating_sub(1)
+                || !self.scroller.can_scroll_down())
```
hrkfdn, line 122: I think this will also trigger pagination if you use the keyboard to go down one line and the screen size is large enough to show all items, meaning it can't scroll down any further:
 
[Screencast from 09-15-2022 10:51:52 PM.webm](https://user-images.githubusercontent.com/19884/190506676-5d75597f-9a66-4877-91db-57583984fcbc.webm)

I can't think of a quick fix though, other than passing whether it was a mouse or keyboard move and preventing pagination in case it was a keyboard move, i.e. `!can_scroll_down() && mouse_event`.

But to be honest the inconvenience of handling this is worth it as it's almost not a problem, unless you have a nicer idea.

hrkfdn (commented):
Hey, thanks a lot for the PR and the added documentation :)

Some comments below.

> There is still a bug present that causes the pagination message to appear at the bottom of the Browse tab, even when there isn't any more items. Help would be wanted with fixing that. 

Unfortunately I haven't found an entry in "Browse" where I can reproduce this. Do you reckon it's because of the change in this PR or a different issue?

src/ui/listview.rs
ThomasFrans: When the page loads on an ultrawide screen for example, it may not show enough items initially to fill the screen. This makes sure that doesn't happen. Maybe there's a better way to solve this...

src/ui/listview.rs
ThomasFrans: Oh I thought it was supposed to load more items when that happened. Unless you mean the case where there are no more items to load, which is handled by the `can_paginate` method I think.

ThomasFrans:
Here is a picture of the Browse category problem (blue text at the bottom of the list shouldn't be there, or should expand, but I think there are no more items to display). Something I noticed is there are exactly 50 items, but when debugging, the paginator says there are 51 items, which may cause the problem I think. I know there are some parts of the Spotify API that can only request 50 items at a time, so this may not be a coincidence that it happens with exactly 50 items. I saw there was a way to tell the pagination to add itself as an extra item, which could cause it to draw the text at the bottom, even when there aren't any more items to display.

image

src/ui/listview.rs
hrkfdn: It's not really a bug, just a change in behavior. But I think it's fine.

hrkfdn:
> Here is a picture of the Browse category problem in the latest release, so not this branch (blue text at the bottom of the list shouldn't be there, or should expand, but I think there are no more items to display). Something I noticed is there are exactly 50 items, but when debugging, the paginator says there are 51 items, which may cause the problem I think. I know there are some parts of the Spotify API that can only request 50 items at a time, so this may not be a coincidence that it happens with exactly 50 items. I saw there was a way to tell the pagination to add itself as an extra item, which could cause it to draw the text at the bottom, even when there aren't any more items to display.

Sometimes Spotify sends a larger page count than can be fetched :( See also https://github.com/hrkfdn/ncspot/issues/131

Unfortunately I can't reproduce it with my "Browse" list

hrkfdn:
Let's get it in!

hrkfdn:
> Unfortunately I can't reproduce it with my "Browse" list

Whoops, I found the culprit. Fixed in 2be1264, thanks for reporting.

ThomasFrans:
Can confirm that Browse works properly now.

PR #972, opened 2022-10-19
fix(todo): Fixed Clone impl marked todo on Pagination

ThomasFrans:
- Fixed a manual implementation of Clone on Pagination (needed extra trait bounds to be derivable)
- Made clippy happy :)


hrkfdn:
Awesome, thanks!

PR #979, opened 2022-10-27
Add new cargo build profiles

ThomasFrans:
I did some tests and with lto and stripping applied (which isn't done for the current release bulids on the Github release page of ncspot), the size of the compiled binary drops from 30M to 13M. Normally package managers strip the binary, but this could be a nice addition for people who download ncspot from the release builds. Since the compilation time almost doubles, I added an optimized profile which is just the previous release profile.

It might make sense to switch these options around, and have something like a "actual-release" profile instead, but then all of the package managers would have to switch to that. I think this is a little more seamless, although now you'd make "normal" release builds with a slightly longer command...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,6 @@ version = "0.11.2"
 # Build time is very slow, but binaries are smaller.
 [profile.release]
 lto = true
-strip = true
 codegen-units = 1
 
 # Used when creating optimized builds for testing.
@@ -23,7 +22,7 @@ codegen-units = 1
 [profile.optimized]
 inherits = "release"
 lto = false
-strip = false
+codegen-units = 16
 
 [badges]
 maintenance = {status = "actively-developed"}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ version = "0.11.2"
 [profile.release]
 lto = true
 strip = true
+codegen-units = 1
 
 # Used when creating optimized builds for testing.
 # Build time is much faster than release, but performance should be about the
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,20 @@ readme = "README.md"
 repository = "https://github.com/hrkfdn/ncspot"
 version = "0.11.2"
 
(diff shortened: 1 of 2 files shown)


hrkfdn:
Hey @ThomasFrans, thanks for the PR. I'm guessing this would also strip function names from backtraces? Or are they missing in release builds anyway? Because I feel like those would be quite valuable as having users run a debug build for a report is a bit of an "entry barrier".

ThomasFrans:
> Hey @ThomasFrans, thanks for the PR. I'm guessing this would also strip function names from backtraces? Or are they missing in release builds anyway? Because I feel like those would be quite valuable as having users run a debug build for a report is a bit of an "entry barrier".

Well... Right now, the backtraces look like this (even without strip, it's smeared out over the terminal) XD
image

But yes, it removes the function names. I didn't think about that. I just noticed a lot of other programs use these settings to get their binary as optimized as possible. I also think most package managers (like Pacman on Arch) strip all the binaries anyway. It doesn't make that much of a difference to leave it off, so I could change it back :)

ThomasFrans:
### Trying to get any/readable backtraces

Out of interest, I was playing around with this for a bit (trying to manually catch panics and restore the terminal before printing a stack trace), but it looks like when the crash occurs after the backend switches to the alternate screen, you don't get a backtrace at all (termion is the only backend that prints anything at all for a stack trace, the other backends don't even do that...). Maybe it would be better to add a crate like [this one](https://docs.rs/backtrace/latest/backtrace/index.html) to try and catch something (it isn't very accurate, but it allows the program to write the backtrace to a file). Doing so would also allow to not print anything to stdout, as this has some weird effects after the backend has been initialized, since it puts the terminal into raw mode, switches screens...

Panic hook:
image

Backtrace in a file:
image

hrkfdn:
Sorry for letting this sit so long. I appreciate your research and investigation into this topic, however I feel like the changes would complicate the build setup a bit which I'd prefer to keep standard. Distributors are free to strip the binary but should consciously do it. That said, the logged backtrace is a great addition! :)

PR #988, opened 2022-11-17
fix(backtrace): Fix backtrace logging and stdout

ThomasFrans:
- Add manual implementation for panic that logs backtrace to a file.
- Remove all manual output to stdout.
- Fix new clippy warnings from Rust 1.65.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -30,9 +30,14 @@ If applicable, add screenshots to help explain your problem.
  - Installed from: [e.g. AUR, brew, cargo]
 
 **Backtrace/Debug log**
-Instructions on how to capture debug logs: https://github.com/hrkfdn/ncspot#usage
+Please attach a debug log and backtrace if ncspot has crashed.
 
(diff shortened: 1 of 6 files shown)


ThomasFrans:
This is a small quality of life commit that tries to fix stdout. Since Cursive controls the tty, ncspot can't reliably print backtraces or other information to it. Therefore it's better to print that information to a file instead. I also removed the message that was displayed at startup, since it also showed some weird behavior (I think it tried to print with a <LF> on linux, but in raw mode, the terminal only moves down, and not to the beginning. After going back out of Raw mode, it doesn't find a valid <CR> and prints that '%' symbol.).

hrkfdn:
Thanks, this looks pretty cool. I have updated the docs and issue template to reflect these changes.
Previously one could redirect `stderr` output to a file to see the backtrace, but this is much nicer.

ThomasFrans:
I saw you added the RUST_BACKTACE=full in the user documentation. This isn't technically required anymore. That option is just there for when you need a backtrace on stderr, but since stderr isn't used for backtraces, you don't need that option. Internally, in the panic handler, it forces a backtrace to be written, even when the user hasn't enabled it explicitly :)

hrkfdn:
Ah thanks! Fixed :)

PR #996, opened 2022-11-26
Make entry point synchronous and switch to global runtime instead.

ThomasFrans:
An example to illustrate what I mean in issue #993. This isn't meant to be merged (therefore a draft PR), but more as a way to show the proposed changes in the issue. Not a lot would have to change. From what I understand about async, this wouldn't have negative performance implications, but I might be wrong about this. I tested this and it works like before.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -117,8 +117,15 @@ struct UserDataInner {
     pub cmd: CommandManager,
 }
 
-#[tokio::main]
-async fn main() -> Result<(), String> {
+lazy_static!(
+    /// The global Tokio runtime for running asynchronous tasks.
+    static ref ASYNC_RUNTIME: tokio::runtime::Runtime = tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .build()
+        .unwrap();
+);
+
+fn main() -> Result<(), String> {
     register_backtrace_panic_handler();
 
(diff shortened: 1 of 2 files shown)


hrkfdn:
So I did some more code (re-)reading and it looks like it makes sense to proceed with this one, as there's barely any async usage starting from `main` (as you also pointed out). We can still revert if that changes. Do you agree or do you want to make more adjustments (as this is still a draft PR)?

ThomasFrans:
I think I noticed two more futures::block_on that I missed as this was just a draft. They also need to be switched. I'll quickly do that, or you can change them. Then I think it's good to merge.

ThomasFrans:
There are still two futures::future::pending() calls, I think somewhere in the `spotify_worker.rs` file. I'm still not too familiar with the difference between futures::* stuff and just the Tokio stuff (like futures::executor vs tokio::runtime). I think it's ok, as I tried without credentials which worked fine, and the pending happens in a bit of code that is responsible for fetching the token, so I think it should work.

hrkfdn:
Those should be fine. `futures::pending()` is just a neverending async task that is non-blocking (i.e. doesn't block a thread using `sleep` or similar). Thanks for the adjustments, hope I'll be able to review and merge this later today.

PR #1009, opened 2022-12-13
Document queue module.

ThomasFrans:
Documentation isn't that necessary for programs, but I like to document while reading/understanding code, and it's so nice to have with the language server to quickly get some info about items, so thought I'd add it to upstream :)


hrkfdn:
Awesome, will happily take it! Thanks!

PR #1011, opened 2022-12-16
Splitting up ListItem and refactoring ListView.

ThomasFrans:
This is absolutely not a final PR. It is meant to get input on this approach. If this would be considered as a change, I would probably make some smaller commits to make it easier to go over the changes, and also improve some parts that are still a bit rough around the edges...

### The interesting parts that changed
`src/ui/list.rs` (replaces `src/ui/listview.rs`)
`src/ui/list/*` (contains all the new list items)

### Reasoning

I have been working on and off on this refactoring of `src/ui/listview.rs` and `src/traits.rs`. It wasn't originally meant to be a PR as it was more for learning how a TUI like cursive worked, but I figured I might as well make a PR out of it. This contains some large changes, which isn't ideal for PR's, but I didn't see a way to split things up, as one change (separating ListItem) had a lot of side effects, and was related to the rewrite of ListView as well. It's still only a learning attempt, and I wouldn't mind if this wasn't considered to be merged. It doesn't add any functionality, it just tries to improve the currently ListView implementation, to make it a bit more readable. I don't even really know if this is a "better" way to do the List, just an attempt :)

### What changed

- List was split into a generic List (like the one in cursive itself, but a bit different), and ListItems.
- ListItem has shrunk and doesn't contain as much stuff anymore. Related things have been put into separate traits in different places, where they make a bit more sense.
- Items that want to be in the list are Views themselves, which makes drawing the list a bit simpler.
- Lots of smaller function signature changes, things I encountered while working on the list and tried to make a bit more general (stuff like Into<Item>, AsRef<Item>, generic items, boxed items).
- No more scroller, ScrollView instead.

### What doesn't work yet
- Pagination for the List.
- Some mouse stuff.
- Not all List functionality has been reimplemented.
- Playing all items in a container is still something that would need some thought.

### Undecided
- Items in the list are kept as data that can be turned into a listitem, not the listitem itself. This makes implementation easier, as the UI changes with the underlying data, but every draw requires the items that are visible to be turned into a Box<dyn ListItem>. I've been keeping an eye on memory and CPU usage, and it doesn't really affect it, but it's a thing that might have to change.


ThomasFrans:
Closing this as I don't believe this is the best approach to this.

PR #1039, opened 2023-01-31
theme config support for cursive's HighlightInactive

han28:
Hi, thanks for this awesome software.

With the color theme I'm using, opening the context menu for a track and tabbing to the "Close" button, it was hard to read the focused (but now inactive) selected item in the list. cursive defaults to using dark blue for the HighlightInactive PaletteStyle for such items in a SelectView, and it seems useful to allow users to override this.

I added a theme option "highlight_inactive_bg" which overrides the default HighlightInactive color from cursive.


hrkfdn:
Merged, thank you!

PR #1053, opened 2023-02-19
let everyone have covers in their notifications

jonas-w:
I don't know why this check exists, because at least to my understanding the cover feature is only for drawing the covers in the terminal. 

I tested this and it displayed the covers in my notifications without needing the "cover" feature, at least i think the cover feature isn't automatically enabled when i compile it with `cargo build`. 

And i personally don't see a reason why covers shouldn't be displayed for everyone. I don't want to use the "cover" feature but having the cover in my notifications is a thing i want.


hrkfdn:
Sounds reasonable, thank you!

PR #1057, opened 2023-02-21
commands: Adding "disconnect" command to force socket shutdown

inemajo:
Hi, this pull request add a "disconnect" command, usefull when your connection change and you want to force closing the socket for reconnect


hrkfdn (commented):
Thanks! Generally this looks good to me. While it is technically more correct to call it `disconnect`, I'm wondering if we should go with `reconnect` connect instead as `disconnect` may suggest the application will close? WDYT?

Also, I haven't really tested this yet, but what is your experience with dropping the connection and have it reconnect. Does ncspot recover from it nicely or is there some extra work needed?

hrkfdn:
Will merge it as `reconnect`, hope that's okay! Thanks again!

inemajo:
Yes it's better to call it reconnect, thanks for the merge !

PR #1059, opened 2023-02-22
Add manpage generation using xtask.

ThomasFrans:
This commit introduces semi-automatic manpage generation using the xtask workflow. This automates generation of manpages using the provided clap Command struct. I only wanted to suggest it for now, so I'm making it a draft.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[alias]
+xtask = "run --package xtask --"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,2 @@
 [alias]
-xtask = "run --package xtask --"
+xtask = "run --quiet --package xtask --"
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@
 tags
 
 /.vscode/
+
+# Ignore generated manpages
+*.1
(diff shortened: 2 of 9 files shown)


hrkfdn:
Hey again! Thanks for the PR, I really like the idea. I'm wondering if there is a way to reuse the Clap `Command` setup from the main project so that we don't have to maintain two instances of it?

Also, I guess the configuration is more complex than the command line arguments, so a manpage for that or having it as part of the ncspot manpage would be cool. I have looked at the `clap_mangen` docs but it seems that this is not possible unless we can somehow add it to the Clap `Command`.

ThomasFrans:
Agreed, I think having a manpage for configuration would be a big improvement. I also thought about having the double instances of the clap command, and I think by making a library for ncspot (src/lib.rs) and putting it in there as a public static or const could help with that. Then both parts of the code could use it as if ncspot was a library.

ThomasFrans:
With this commit, I think the man page generation using xtask could be ready. It doesn't include the configuration information as I think it would make more sense to provide this information the same way Alacritty and some other programs do, using a default configuration file with added comments that explain the purpose of each option (I could make a separate PR for that). If adding the configuration to the man page would be preferred over the 'default configuration' approach, I'll happily add it.

I also added a bit of info about the provided metadata files to the README for packaging ncspot. I hope that's ok.

hrkfdn:
Very cool, merged. Thanks!

PR #1060, opened 2023-02-23
Fix queueview shuffle order.

### LINKED ISSUE #465: Show playing order in queue
When playing a playlist on shuffle, the queue page is just a copy of the playlist in order, not showing the actual order the songs will be played in. Is there a way to see this?

ThomasFrans:
This bugfix changes the way the `ListView` renders the content. It adds a way to add an explicit order of items to `ListView` in the same way as `Queue` has, and uses that for rendering. I just want to ask to quickly check this one to make sure it doesn't break other things (I don't think adding the order and using it only for rendering should have side effects, but I'm not 100% sure about it).

closes #465

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -120,7 +120,13 @@ impl CommandManager {
                         self.queue.get_current_index()
                     );
                     s.queuestate.queue = queue.clone();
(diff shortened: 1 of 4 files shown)


hrkfdn:
Thanks a lot! LGTM, merged!

PR #1062, opened 2023-02-23
CI: add clippy and update toolchain install action.

### LINKED ISSUE #1051: Prevent clippy warnings by integrating into CI.
I was wondering whether there is a way to prevent accidental commits that aren't free of clippy warnings by having clippy check the code in CI. This would obviously not help against updates to the clippy database of checks if there was already code with those errors in the codebase, but it could at least notify that there are errors in new commits. Maybe it's an idea to work with a set toolchain, although that would force all contributors to download the toolchain next to their stable version.

In a personal project I just added clippy to CI with the following YAML and from then on it would prevent any more commits that already contain faulty code.

```yaml
      - name: Check Code Quality
        run: cargo clippy --all-targets -- -D warnings
```

There could be better ways to do this. Ultimately I'm just proposing to find a way to prevent clippy warnings from being committed as that keeps the codebase cleaner and integrating it into CI somehow would enforce that. It would prevent those small "fix clippy warnings" commits :p

ThomasFrans:
- Update all the actions to the latest version since some of them were using deprecated stuff.
- Use a different Rust toolchain installation action because the previous one is no longer maintained and generated deprecated warnings. The new one is by dtolnay (the author of the syn, anyhow and cargo-expand) so it will probably be maintained for a while :slightly_smiling_face: 
- Fix the last clippy warning that occurred by enabling all features.
- Cache the build artifacts as long as the dependencies don't change.
- Build the project instead of just checking it (I'm not sure but I think just checking can miss errors under some circumstances. Some projects check, others build.). Building doesn't add much extra time when caching the `target` directory anyways.

I hope this is ok. This is only my second GitHub CI config I've worked on so if anything isn't right, I'll happily change it :slightly_smiling_face:

The first run without any caches took around 10m for me, subsequent runs take around 3m40s as long as the `Cargo.lock` file doesn't change (which invalidates the dependency artifact caches).

Also it might be necessary to manually delete the latest caches (the ones that were used in the last while) as I don't think they will be updated with the target directory otherwise...

closes #1051

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,8 @@ on:
     branches: [ main ]
 
 jobs:
-  check-macos:
+  # Build the Rust code.
+  build:
     name: Checking ${{ matrix.build_target }}
     runs-on: ${{ matrix.os }}
     strategy:
@@ -30,13 +31,21 @@ jobs:
             target: x86_64-pc-windows-msvc
             features: '--no-default-features --features rodio_backend,pancurses_backend,share_clipboard,notify'
     steps:
-      - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
+      - uses: actions/checkout@v3
+      - uses: actions/cache@v3
+        name: Cache build data
         with:
-          toolchain: stable
-          override: true
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            ~/.cargo/bin
+            target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
           target: ${{ matrix.target }}
-          profile: minimal
+          components: clippy, rustfmt
       - name: Install macOS dependencies
         if: matrix.os == 'macos-latest'
         run: brew install portaudio pkg-config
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -45,23 +54,48 @@ jobs:
...
-            ~/.cargo/registry
-            ~/.cargo/git
-          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-      - name: Running cargo check
-        uses: actions-rs/cargo@v1
-        with:
-          command: check
-          args: --locked --release --target ${{ matrix.target }} ${{ matrix.features }}
+      - name: Running cargo build
+        run: cargo build --locked --target ${{ matrix.target }} ${{ matrix.features }}
```
hrkfdn, line 58: Is there a reason you went for `cargo build` instead of `cargo check`?

It seems that `cargo check` has some advantages in that it is a tad faster: https://doc.rust-lang.org/cargo/commands/cargo-check.html

.github/workflows/ci.yml
```diff
@@ -30,13 +31,21 @@ jobs:
...
     steps:
+      - uses: actions/checkout@v3
+      - uses: actions/cache@v3
+        name: Cache build data
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            ~/.cargo/bin
+            target
```
hrkfdn, line 42:
```suggestion
            ~/.cargo/bin/
            ~/.cargo/registry/index/
            ~/.cargo/registry/cache/
            ~/.cargo/git/db/
            target/
```
Ah, thanks for taking another look at this. This is the path list currently listed in the `actions/cache` examples. Maybe it's more specific?

.github/workflows/ci.yml
```diff
@@ -45,23 +54,48 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/cache@v3
+        name: Cache build data
+        with:
+          path: |
+            ~/.cargo/registry
+            ~/.cargo/git
+            ~/.cargo/bin
+            target
```
hrkfdn, line 88:
```suggestion
            ~/.cargo/bin/
            ~/.cargo/registry/index/
            ~/.cargo/registry/cache/
            ~/.cargo/git/db/
            target/
```
I guess these would also need to be adjusted (if we go for it).

hrkfdn (commented):
Great! Thanks for tackling this. I have some minor comments below.

.github/workflows/ci.yml
ThomasFrans: Cargo check is significantly faster than cargo build on the first run. After saving the cache, cargo build isn't that much worse. The documentation mentions that there are some errors cargo check can't detect. I thought for correctness cargo build could be better. Either one would work decently though.

.github/workflows/ci.yml
ThomasFrans: I don't know what is in the registry, or where I found that path exactly. It worked fine so I think it's ok, but maybe being more specific could help prevent problems by keeping files that could cause problems with toolchain/package updates out of the cache. Either one could be tested for a bit and if it doesn't work, the other one can be used 🙂

ThomasFrans:
I'll update the things you mentioned (or you can do it if you prefer).

.github/workflows/ci.yml
hrkfdn: I'll just commit them, I'd assume since they're listed in the `actions/cache` docs they should be fine

hrkfdn (approved):
Merged, thank you! :)

ThomasFrans:
> Merged, thank you! :)

I think you might have to clear out the current job cache once (the ones used in the last while, as I don't remember if there is a clear all button). I had some problems where the clippy and build files were using the same cache tag which isn't correct.

hrkfdn:
Thanks for the heads up, I'll update the Cargo.lock via a dependency update which should change the cache key, right?

PR #1068, opened 2023-03-02
Automatic shell completion generation using xtask.

ThomasFrans:
This adds the option to automatically generate shell completion scripts for a variety of shells (currently bash, zsh, fish, elvish and powershell) using `cargo xtask generate-shell-completion`. It uses the crate `clap_complete` to automatically create these completions based on the `clap::Command` struct. It could be handy for package maintainers to add these scripts to their package.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[alias]
+xtask = "run --package xtask --"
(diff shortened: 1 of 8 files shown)


ThomasFrans:
I feel like I'm not contributing a lot of code in the last few PRs...
...

Anyways, automatic shell completion script generation using xtask. Now we never have to type `--debug` again, because it's so long... Totally worth it... :neutral_face:

I know this isn't really required but I like when apps adhere to the most common conventions/rules of an OS, so I thought being able to generate shell completions would be nice for package maintainers, so they can add shell completion scripts with the package. What do you think? 😁

Next up: Appstream?

PR #1069, opened 2023-03-02
Provide upstream Appstream file.

ThomasFrans:
Add an Appstream file for the ncspot program for representation in Linux app stores. An Appstream file provides packagers that package ncspot for Linux app stores with the necessary metadata about the program. The file is supposed to be provided upstream instead of by packagers.

### DIFF
--- a/misc/com.github.ncspot.metainfo.xml
+++ b/misc/com.github.ncspot.metainfo.xml
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="console-application">
+  <id>com.github.hrkfdn.ncspot</id>
+  
+  <name>ncspot</name>
+  <summary>Play Spotify music from the terminal</summary>
+  
+  <metadata_license>CC0-1.0</metadata_license>
+  <project_license>MIT</project_license>
+  
+  <description>
+    <p>
+      ncspot is an ncurses Spotify client written in Rust using librespot. It is heavily inspired by ncurses MPD clients, such as ncmpc.
+    </p>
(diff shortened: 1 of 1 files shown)


ThomasFrans:
Since ncspot is being packaged for Flatpak, I thought it made sense to properly support it upstream. Flathub makes use of the [Appstream metadata format](https://www.freedesktop.org/software/appstream/docs/) which is used by many Linux app stores and pretty much the standard for providing application metadata for that purpose. Appstream dictates that the metainfo file should be provided upstream. This is currently not the case as the Flathub manifest provides it downstream. Providing it upstream would guarantee consistent details across app stores and would help package maintainers who wouldn't have to individually maintain the file.

This is still a draft for two reasons. First of all, this is more of a feature request with added context (the initial commit towards adding Appstream support upstream). The file is currently only the information that can be generated by the Freedesktop Appstream generator, which does a decent job, but doesn't include more detailed information. If this PR is a welcome addition, I would fill out the more detailed information.

The second reason is because I think in the spirit of automation (because we're developers, and we're lazy), I think it makes sense to automate updates to the file (version, changelog maybe...). Many projects do this by using a separate build system (Meson in many cases), but I like the vanilla cargo experience in Rust, so I thought maybe it would be handy to update the file with (yet) another xtask that uses information from other places (the Cargo.toml for the version, a separate json/ron file for structured changelogs) and puts that in the metainfo file, which could be a (Mustache) template and can be generated upon release. None of this would increase the size of ncspot of course, it would just be a separate build step for packagers that want/need the Appstream info.

PR #1070, opened 2023-03-03
Disable autojump for add to playlist dialog.

### LINKED ISSUE #881: Possibly a new regression in contextmenu navigation
**Describe the bug**
Hello, I noticed that when adding a track to a playlist, pressing `j` will jump between all playlists starting with the letter j, and not going down as expected. I can't confirm it happens with pressing `k`, as I don't have playlists starting with that letter.

**To Reproduce**
1. Open a track's contextmenu by pressing `o`
2. Choose "Add to playlist"
3. Press `j`
4. Observe jumping between playlists starting with the letter j

**Expected behavior**
Navigating up and down with `j` and `k` should work the same way as with up and down arrow keys.

 - OS: Arch Linux
 - Terminal: Alacritty
 - Version: `master` branch
 - Installed from: AUR & cargo

**Additional context**
This issue was solved back in #335, but the issue seems to be back.

ThomasFrans:
Autojump seems to take precedence over regular keys, which isn't the wanted behavior, especially when there are playlists in the list that contain some of the most common Vim keybindings. For example when there is a playlist called 'jazz', pressing `j` will jump to that playlist instead of going down.

This commit disables autojump for now. If search inside the add to playlist dialog would be a wanted feature, it could be implemented using the default Vim search button `/` which would work the same way as searching in other lists works right now.

closes #881

### DIFF
--- a/src/ui/contextmenu.rs
+++ b/src/ui/contextmenu.rs
@@ -67,7 +67,6 @@ impl ContextMenu {
             }
         }
 
-        list_select.set_autojump(true);
         list_select.set_on_submit(move |s, selected| {
             let track = track.clone();
             let mut playlist = selected.clone();


hrkfdn:
Merged, thanks!

PR #1071, opened 2023-03-03
Change unnecessary usage of Arc to borrow instead.

ThomasFrans:
Some basic cleanup of function signatures that took ownership of their parameters, even though they didn't need ownership. Switching over the usage of `Arc` to a normal borrow has the added benefit of cleaning up the code a bit since now a reference can be given instead of having to clone the values. The other benefit is that a lot of clones aren't necessary anymore. It's not going to have noticable performance benefits, but it is still a good thing to have less clones all over the code.

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -52,7 +52,7 @@ impl CommandManager {
         config: Arc<Config>,
         events: EventManager,
     ) -> CommandManager {
-        let bindings = RefCell::new(Self::get_bindings(config.clone()));
+        let bindings = RefCell::new(Self::get_bindings(&config));
         CommandManager {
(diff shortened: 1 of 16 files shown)


ThomasFrans:
I hope this sort of general cleanup is accepted. I try not to do it for small stuff as it's probably annoying to have 100 small PRs with general cleanup. I only did this one as it also makes the code more understandable in my opinion as now it is clear that all those functions don't take ownership of their values. If this isn't ok, let me know :slightly_smiling_face:

hrkfdn:
Very nice, thank you! Just a few days ago I wondering whether this could be improved.

hrkfdn (approved):

PR #1074, opened 2023-03-04
Emulate double click to play items.

### LINKED ISSUE #917: [Mouse Support] Middle click to play an item, left click to open it
This is just a very short request that's related to #840 that was closed some weeks ago. The new mouse support works really well, but the one feature I really wanted (being able to play songs with the mouse) wasn't implemented. I absolutely understand why, I also don't like it at all when you click a song by accident and your queue gets cleared, so I was wondering if it's possible to have a more deliberate action that can't be done by accident to play songs. I tried using middle click to play items in a ListView, and left click to open them (if possible). Middle click might not make the most sense, but it isn't used for anything at the moment, and middle click is often used to open links in other programs (Firefox, File explorers...).

Left click -> Open an item if possible
Middle click -> Play an item if possible (doesn't work right now for categories, just clears the queue which isn't really expected, `play` for the category `ListItem` could play the first `Playlist` inside it)
Right click -> Open context menu

In `ListView.rs` (idk if this is correct code, it just works as a prove of concept rn :))

```
            Event::Mouse {
                event: MouseEvent::Press(MouseButton::Middle),
                position,
                offset,
            } => {
                let viewport = self.scroller.content_viewport().top_left();
                let selected_row = position.checked_sub(offset).map(|p| p.y + viewport.y);
                if let Some(y) = selected_row.filter(|row| row < &self.content_len(false)) {
                    self.move_focus_to(y);

                    if let Some(_) = {
                        let content = self.content.read().unwrap();
                        content.get(self.selected).map(|t| t.as_listitem())
                    } {
                        self.queue.clear();

                        if !self.attempt_play_all_tracks() { let mut content = self.content.write().unwrap();
                            if let Some(item) = content.get_mut(self.selected) {
                                item.play(self.queue.clone());
                            }
                        }

                        return EventResult::Consumed(None);
                    }
                }
                
            }
```

ThomasFrans:
Change the behavior of `ListView` to treat a click on an item that is already selected as a double click. It allows the user to play songs (not episodes since currently it's not possible to determine if a `ListItem` is an episode) by "double clicking" them. It is a bit of a hack, but it works pretty well. A possible downside is that when people that don't like mouse integration in a TUI click to focus the terminal window, it could jump to the song they clicked if that happened to be the selected one.

closes #917

### DIFF
--- a/src/ui/listview.rs
+++ b/src/ui/listview.rs
@@ -132,6 +132,11 @@ impl<I: ListItem + Clone> ListView<I> {
         }
     }
 
+    /// Return whether this [ListView] has visible scrollbars.
+    pub fn has_visible_scrollbars(&self) -> bool {
+        self.scroller.get_show_scrollbars()
+    }
+
     pub fn get_selected_index(&self) -> usize {
         self.selected
     }
@@ -161,6 +166,9 @@ impl<I: ListItem + Clone> ListView<I> {
         self.move_focus_to(max(new, 0) as usize);
     }
 
+    /// Append the currently selected item and all the following ones to the queue after the
+    /// currently playing track and start playing them. Returns true if adding and playing the
+    /// tracks succeeded, false otherwhise.
     fn attempt_play_all_tracks(&self) -> bool {
         let content = self.content.read().unwrap();
         let any = &(*content) as &dyn std::any::Any;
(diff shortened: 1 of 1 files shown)


ThomasFrans:
I don't know if this is an acceptable approach to implement this feature request, as there is the problem explained in the commit message. I don't think this is too bad of a problem personally since clicking anywhere already highlights songs. A configuration option to disable the mouse could be added for people that prefer no mouse integration in TUI applications.

hrkfdn:
Merged, thanks! :)

PR #1079, opened 2023-03-06
Fix debian dependencies for Bookworm

treibholz:
Bookworm favours pipewire over pulseaudio, so the dependency to the package "pulseaudio" from my previous PR #405 has to be removed. It seemes, it was never really necessary anyway, because "$auto" already adds the correct dependency to "libpulse0".


hrkfdn:
Thank you! :)

PR #1080, opened 2023-03-07
Automatic shell completion generation using xtask.

ThomasFrans:
This adds the option to automatically generate shell completion scripts for a variety of shells (currently bash, zsh, fish, elvish and powershell) using `cargo xtask generate-shell-completion`. It uses the crate `clap_complete` to automatically create these completions based on the `clap::Command` struct. It could be handy for package maintainers to add these scripts to their package.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,10 @@ tags
 
 /.vscode/
 
-# Ignore generated manpages
+# Ignore generated resources
 *.1
+misc/_ncspot
+misc/ncspot.bash
+misc/ncspot.fish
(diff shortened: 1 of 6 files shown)


ThomasFrans:
Quickly remade this PR as I based it on the work I did for the man page generation since this also required the xtasks. I don't know Git well enough to know what demons merging that previous one would have unleashed, so I quickly redid it based on the main branch which now has the xtask stuff included.

hrkfdn:
Merged, thanks!

Issue #1061, opened 2023-02-23
Shrink README to most important items, move documentation/manual.


ThomasFrans:
**Is your feature request related to a problem? Please describe.**
I was browsing through the README today to find some information I needed to work on a bugfix and found the current one a bit hard to navigate and a bit overwhelming. I realize the README is the Git repo's entrypoint, the `fn main()` where new users land and current users expect to find the most important info so I'm a bit scared to suggest changes to it, especially since I think the current one has been the same for a long time.

**Describe the solution you'd like**
What I would want to suggest is to shrink down the current README to only contain the essential stuff, and provide links to resources that group together related information. The current one has a table of contents but it's a bit long and I think no user would need all the information in there at once. I think therefore it could make sense to split up that info into separate pages/a new documentation format (like mdbook). There could be categories of information, since a user would most likely want to get to information regarding one specific topic as quickly as possible:
- development info
- packaging info
- installation info
- configuration info
- contribution guidelines
- license
- issues
- ....
- 
A user falls into roughly two categories: 
- A new user, who wants to get a quick glimpse of what the project is about and how to install it/where to get it. For them it makes sense to offer that info straight in the README as they don't want to have to click around to different pages to get the information.
- A current user/developer, who knows the project and is looking for specific information in one of the aforementioned categories.

**Describe alternatives you've considered**
I've considered 3 ways to make the current README a bit more clear:
- Create a GitHub wiki page for the items that are currently in the README, but are meant for current users/developers. Keep the project description and installation instructions in the README and mention the wiki for extra information. The disadvantage of this approach is that the project becomes dependent on GitHub, which I personally try to avoid.
- Create separate `*.md` files that contain information related to a specific topic. Links to them could be provided cleanly in the main README file, and each sub-file could contain its own smaller table of contents.
- Create mdbook documentation for the project. I don't know if this is the best approach as it adds complexity and there isn't that much information that it needs a whole documentation page.

**Additional context**
If a suggestion would be wanted, I could create a PR to show what I was thinking about. I know the README of a project is very important and I don't want to change it unnecessarily, but I feel like splitting it up could make it easier to quickly get to the relevant information. I could make a PR with a suggestion if it would help :slightly_smiling_face:. It would probably roughly put the current top-level items into separate files.

**Examples**
- [Alacritty's README](https://github.com/alacritty/alacritty#readme)
- [Neovim's README](https://github.com/neovim/neovim#readme) (contains a bit more info, but still links to other places regarding extra information)

[...]

PR #1089, opened 2023-03-16
Update README.

ThomasFrans:
This is just an example of the change I proposed in #1061. I'm not an artist and my creative skills are about as good as a toddler (maybe even worse). This is just an attempt to show what I meant by shrinking the main README and providing information for users, developers and packagers separately. Therefore the 'subpages' aren't finished yet. I wanted to focus on the main README and ask input first.

Some things I tried to do:
- Increase the size of the screenshot and put it closer to the start.
- Divide the main README into categories based on each type of user, so a user can quickly find installation and configuration info, a developer can find build instructions and a package maintainer can find packaging info.
- Put the most important information for users at the top of the README.
- Remove the long table of contents in favor of a shorter README.


hrkfdn:
Good call, the README has become quite cluttered. Some comments:

- The main page is very clean, love it!
- The packaging statistics image is quite large, but I think it goes well with the installation instructions. As a user of a specific distro it may be helpful to know it's in the package repositories
- Maybe the building and packaging docs can be merged as they accompany each other quite well. Or do you think that would blow it up too much again?

ThomasFrans:
> The main page is very clean, love it!

I'll be honest, I got the idea for the top section from browsing to other GitHub pages. I really liked the way the [Helix README](https://github.com/helix-editor/helix#readme) looked, so I got (a lot of) inspiration from their README. One problem with this README is that the icon doesn't show on the GitHub app, as there seems to be a bit of a problem there with parsing the HTML for switching out the icons. It just shows the alt text. A solution could be to use a gray-ish color for the text and not to use the switching behavior for light/dark.

> The packaging statistics image is quite large, but I think it goes well with the installation instructions. As a user of a specific distro it may be helpful to know it's in the package repositories

I didn't think about that. I thought it would be good for packagers to know which platforms already have a package available, but indeed it also makes a lot of sense in the installation instructions. Like I mentioned I didn't spend much time on the subcategories yet so there is still a bit of work left there (reordering things, moving/copying things between them...).

> Maybe the building and packaging docs can be merged as they accompany each other quite well. Or do you think that would blow it up too much again?

I split them because linking to another file on GitHub doesn't take sections into account (developers.md#section), that only works for the page you're currently on. So for a packager, they would click 'more info' in the main README and it would take them to the top of `developers.md`, not to the packaging section. I think it makes sense that they also just need the build instructions and the extra packaging info so it might be ok to include it and just have them land on the build instructions first. Another option would be to explain the process of building (in general, no details) and packaging in the packaging file, and link to the detailed instructions from there, although that's yet another link to somewhere else, and too much clicking to get everywhere also becomes annoying. I'll play around with it for a bit :slightly_smiling_face:

ThomasFrans:
I added the package information to the user installation instructions and also did some other small changes to make sure everything is up to date and correct. Most of the configuration section is just copied over to `doc/users.md` for now as I think it makes sense to provide a file with the default options for that in the future, and shrink the section in the README a bit further.

Regarding the packaging section, I considered adding it to the `doc/developers.md` file but I think they really don't fit together as they are giving information about two completely different things. The developer page gives development information and gives commands to create development builds, while the packaging page gives instructions on how to compile a release build. I feel like it really makes sense to split them as not every packager might necessarily  be a developer. That's why I included specific build instructions for the packaging page and provided a link to the developer instructions in case more information is absolutely required.

The version of the packaging page should now give all/only the commands required to build a package, while the development page provides all/only the information to develop ncspot. If this approach isn't ok, I'll happily move the packaging stuff over to the development page :slightly_smiling_face:

hrkfdn:
Merged! The reason I thought they could go together is because for instance the feature flags may be relevant for packagers, but since there is a link that also works.

Thanks a lot for cleaning up the docs!

jonathannerat:
@ThomasFrans sorry for the bump, do you remember what font is used in the main screenshot? I couldn't find it in the README nor here.

ThomasFrans:
@jonathannerat I don't know what font it is. I just copied the screenshot when changing the README as I didn't want to change the actual content too much.

Issue #824, opened 2022-06-04
`use_nerdfont` broken on MacOS


isinyaaa:
**Describe the bug**
Enabling the option with nerdfonts installed does not work properly

**To Reproduce**
1. Install nerdfonts in your system
2. Enable the setting on ncspot
3. Open ncspot
4. Look at status bar icons

**Expected behavior**
Nerdfonts should work

**Screenshots**
<img width="1501" alt="image" src="https://user-images.githubusercontent.com/39812919/172026527-495a219a-fc17-4447-ae02-532d6d5aed2c.png">

**System (please complete the following information):**
 - OS: [e.g. Linux, macOS]
 - Terminal: [e.g. GNOME Terminal, Alacritty]
 - Version: [e.g. 0.4.0, `master` branch]
 - Installed from: [e.g. AUR, brew, cargo]

**Backtrace/Debug log**

<details><summary>stderr.log</summary>
<p>

```raw
warning: unused import: `error`
 --> src/queue.rs:5:18
  |
5 | use log::{debug, error, info};
  |                  ^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: field is never read: `notification_id`
  --> src/queue.rs:41:5
   |
41 |     notification_id: Arc<AtomicU32>,
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: function is never used: `cache_path_for_url`
  --> src/utils.rs:23:8
   |
23 | pub fn cache_path_for_url(url: String) -> std::path::PathBuf {
   |        ^^^^^^^^^^^^^^^^^^

warning: function is never used: `download`
  --> src/utils.rs:29:8
   |
29 | pub fn download(url: String, path: std::path::PathBuf) -> Result<(), std::io::Error> {
   |        ^^^^^^^^

warning: associated function is never used: `cover_url`
  --> src/model/playable.rs:85:12
   |
85 |     pub fn cover_url(&self) -> Option<String> {
   |            ^^^^^^^^^

warning: `ncspot` (bin "ncspot") generated 5 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 0.28s
     Running `target/debug/ncspot -d debug.log`
```

</p>
</details>

<details><summary>debug.log</summary>
<p>

```raw
[2022-06-04][17:15:55] [ncspot] [INFO] Using cached credentials
[...]

PR #1093, opened 2023-03-18
statusbar: update nerdfont glyphs

cfebs:
As of https://github.com/ryanoasis/nerd-fonts/issues/1096 nf-mdi* glyphs are marked obsolete in nerd-fonts upstream

Update the glyphs to non-deprecated nf-fa* counterparts

## Details

The current glphys used in ncspot are marked obsolete:

| play | pause | stop |
|--------|--------|--------|
| image  | image | image | 

> From https://www.nerdfonts.com/cheat-sheet

An added bonus to the update may be a bugfix too! There is an issue opened here: https://github.com/hrkfdn/ncspot/issues/824 which is similar to my experience.

Here is what the glyphs look like for me just echoing the unicode (kitty 0.27.1, `TERM=xterm-kitty`, `font_family JetBrains Mono`, [my installed nerd-font packages][1])

`echo -e 'CURRENT:\n\uf909\n\uf8e3\n\uf9da\nNEW:\n\uf04b\n\uf04c\n\uf04d'`

image

and here is a comparison from my "current" ncspot distro (arch) ncspot (0.12.0) vs. "new" local release build:

#### current

image

#### new

image

[1]: https://gist.github.com/cfebs/3c06bd89e5854a3dcd4841b0cd32ec35

### DIFF
--- a/src/ui/statusbar.rs
+++ b/src/ui/statusbar.rs
@@ -47,9 +47,9 @@ impl StatusBar {
             .flip_status_indicators
             .unwrap_or(false);
 
-        const NF_PLAY: &str = "\u{f909} ";
-        const NF_PAUSE: &str = "\u{f8e3} ";
-        const NF_STOP: &str = "\u{f9da} ";
+        const NF_PLAY: &str = "\u{f04b} ";
+        const NF_PAUSE: &str = "\u{f04c} ";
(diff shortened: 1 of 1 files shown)


hrkfdn (approved):
Thanks a lot! LGTM!

PR #1108, opened 2023-04-01
Add lyrics page

### LINKED ISSUE #1076: Add time-synced lyrics tab
**Describe the solution you'd like:**

View lyrics like in official app. Bonus points if time-synced, but no so necessary. I can make this myself using author+song_name from dbus, but would be highly appreciated to see this as native feature.

Naapperas:
This PR adds a new screen for the user to see the currently playing track's lyrics and optionally save them to their system.
Closes #1076.

This behavior should be configurable by the user, i.e.:
- the location for saving lyrics should be configurable (**WIP**);
- the users preferred fetcher(s) should be configurable. For instance, they may want to only use [MusixMatch](https://www.musixmatch.com/) or they may want to have several options available, with the next fetcher being used in case the previous is unavailable/invalid (**WIP**).

Missing features:
- Lyric fetching (need to think about how to correctly abstract the fetching mechanism to use several implementations)
- Lyrics page update (no access to music change event, at least none I could find, need to "manually" update, aka re-enter the view)
- Asynchronous lyrics fetching (I looked into [Cursive Async View](https://docs.rs/cursive-async-view/latest/cursive_async_view/) but could not get it to work, still need to investigate further)

Still a Work In Progress but any comments/reviews/help are appreciated.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1872,7 +1872,7 @@ dependencies = [
 
 [[package]]
 name = "ncspot"
-version = "0.13.0"
+version = "0.13.1"
 dependencies = [
  "chrono",
  "clap",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ license = "BSD-2-Clause"
 name = "ncspot"
 readme = "README.md"
 repository = "https://github.com/hrkfdn/ncspot"
-version = "0.13.0"
+version = "0.13.1"
 
 [badges]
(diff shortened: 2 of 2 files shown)


Naapperas:
Progress is slow due to college work but this is not forgotten

hrkfdn:
@Naapperas No worries, good luck with college :)

Naapperas:
I was trying to fix some organizational aspects of the PR and i guess Github closed it for me (?!?!)

PR #1127, opened 2023-04-23
Use notify-rust with zbus instead of dbus and add ncurses_backend feature flag

jirutka:
Please see commit messages.


hrkfdn:
Good catch, thanks!

PR #1150, opened 2023-05-06
Fix incorrect notification ID reuse on XDG

cyqsimon:
Previously only the first notification (for each time the application launches) is actually sent on XDG (Linux) because the notification ID was reused incorrectly. This is a simple fix that removes the reuse.

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -509,13 +509,8 @@ pub fn send_notification(
     cover_url: Option<String>,
     notification_id: Arc<AtomicU32>,
 ) {
-    let current_notification_id = notification_id.load(std::sync::atomic::Ordering::Relaxed);
-
     let mut n = Notification::new();
-    n.appname("ncspot")
-        .id(current_notification_id)
-        .summary(summary_txt)
-        .body(body_txt);
+    n.appname("ncspot").summary(summary_txt).body(body_txt);
 
     // album cover image
     if let Some(u) = cover_url {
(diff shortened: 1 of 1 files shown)


hrkfdn:
Hey, thanks for the PR.

I currently only have access to a macOS system so can't investigate this, but reading from the code this removes the ID reusage completely, right? So whenever a user skips through a bunch of tracks it will always generate a new notification. Was the old behavior causing problems? It worked quite nicely on my GNOME setup.

cyqsimon:
> It worked quite nicely on my GNOME setup.

It's not working too well on my Arch KDE box. The first time a new song starts playing after ncspot startup, a new notification is shown as expected and goes away after a few seconds (the default timeout configured by the user). This is all good. But since the notification is set to transient, it does not stay in the notification widget, so basically it's just gone. Then afterwards, an attempt to reuse the same notification id simply does nothing.

I think we may have different interpretations and expectations of what this feature is supposed to do. Personally, I do expect a new notification for every new song and I'm happy with it. I'm completely guessing, but maybe what you expect is to have a persistent notification that stays in the notification widget/area? (And it's working as expected for you because maybe Gnome doesn't respect the "transient" hint?) Please clarify.

hrkfdn:
I see how that's annoying. In GNOME the notification will stay in the "notification panel" until it is explicitly dismissed. So for every track change the notification will persist if the ID is not reused. But maybe notifications aren't really used on GNOME as it supports MPRIS out of the box, anyway.

hrkfdn:
So it seems GNOME is different with this, but I think your behavior is correct. I have added a small change that removes the notification ID storage as it's unused.

hrkfdn (approved):

PR #1153, opened 2023-05-06
Bump rspotify from 0.11.6 to 0.11.7 and use native TLS only

jirutka:
reqwest uses native TLS by default, but rspotify with ureq currently uses rustls.


hrkfdn:
Merged, thank you!

PR #1163, opened 2023-05-11
fix(ui): ListView decrement selected on remove

ThomasFrans:
When removing an element, it is important to check whether it was the last element in the list and decrement the selected index if so.


hrkfdn:
Merged, thanks!

PR #1176, opened 2023-05-28
Prevent tab switching from the command line

### LINKED ISSUE #1133: Scrolling tabs when searching items
**Describe the bug**
When the user presses `/` (thus enabling the search bar) they can type characters through which they can navigate. However, once reaching the "edges" of the search bar buffer, further presses instead switch between tabs.

**To Reproduce**
1. Press `/`
2. Type anything (I believe this happens without anything on the buffer but the common case is when users are typing)
3. Press the right arrow button
4. See the tabs being switched

**Expected behavior**
Once we go to either end of the text on the search bar, further attempts to go pass that do nothing.

 - OS: EndeavorOS
 - Terminal: xfce-terminal
 - Version: 0.13.0-1
 - Installed from: AUR

ThomasFrans:
When moving inside the command line, if the cursor is at either edge of the buffer, the events aren't consumed and propagated further into the view tree. This prevents that behavior.

fixes #1133

### DIFF
--- a/src/ui/layout.rs
+++ b/src/ui/layout.rs
@@ -247,6 +247,22 @@ impl Layout {
             None
         }
     }
+
+    /// Propagate the given event to the command line.
+    fn command_line_handle_event(&mut self, event: Event) -> EventResult {
+        let is_left_right_event = matches!(event, Event::Key(Key::Left) | Event::Key(Key::Right));
+        let result = self.cmdline.on_event(event);
+
+        if self.cmdline.get_content().is_empty() {
+            self.clear_cmdline();
+        }
+
+        if is_left_right_event {
+            EventResult::Consumed(None)
+        } else {
+            result
+        }
+    }
 }
 
 impl View for Layout {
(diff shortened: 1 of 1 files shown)


src/ui/layout.rs
```diff
@@ -247,6 +247,22 @@ impl Layout {
             None
         }
     }
+
+    /// Propagate the given event to the command line.
+    fn command_line_handle_event(&mut self, event: Event) -> EventResult {
+        let is_left_right_event = matches!(event, Event::Key(Key::Left) | Event::Key(Key::Right));
+        let result = self.cmdline.on_event(event);
```
hrkfdn, line 254: Nit: It probably works fine because `result` contains a callback, but I wonder if the call to `self.cmdline.on_event()` should be moved inside the `else` block? Otherwise we pass events to the `cmdline` even if it's a left/right keypress.

hrkfdn (approved):
Looks good, one minor comment below. Thanks!

src/ui/layout.rs
ThomasFrans: I think I understand what you mean, but I don't know whether it would be correct. If `self.cmdline.on_event(event)` was moved to the else, the user couldn't move left and right anymore inside the command line.

hrkfdn: Ah yes, you're right.

PR #1182, opened 2023-06-03
fix: command line multi-byte prefix panic

ThomasFrans:
Don't index into the command line command as the prefix can be altered by the user and therefore be a multi-byte unicode character.


ThomasFrans:
I was aware of this bug already but didn't submit a patch as I didn't think users could edit the command. I forgot that users can move 'in front of the prefix', which can cause a crash when they submit a command that contains a multi-byte character as the prefix, like '🔥' for example.

PR #1185, opened 2023-06-04
fix: config option `command_key` not working

ThomasFrans:


ThomasFrans:
I had made a mistake when pattern matching the event in the layout in #1172. This is a quick fix for that. I should have double checked this.

I find it a bit weird how the order of the match matters. Maybe it would be better to not match on `Event::Char(character)` and instead on `Event::Char(character) if self.cmdline_focus && (character == '/' || character == self.configuration.values().command_key.unwrap_or(':'))`. It's way longer but it would prevent the order of items in the match expression from causing confusing behavior when the next person is working on it.

I don't know what to choose between smaller match patterns for readability (the way it is in this PR) or more precise ones. Maybe there are general guidelines for which one to prefer in the Rust documentation somewhere, but I'm not aware of any.

hrkfdn:
> I find it a bit weird how the order of the match matters. Maybe it would be better to not match on `Event::Char(character)` and instead on `Event::Char(character) if self.cmdline_focus && (character == '/' || character == self.configuration.values().command_key.unwrap_or(':'))`. It's way longer but it would prevent the order of items in the match expression from causing confusing behavior when the next person is working on it.

Agreed. I'd say that if changing the order of match conditions changes behavior that's too implicit and would get lost over time.

hrkfdn:
Merged, thanks for adjusting!

ThomasFrans:
I hope that was the only oversight. If there are more bugs/regressions because of the refactor, I'll happily fix them :slightly_smiling_face:

PR #1195, opened 2023-06-06
docs: add `cargo` installation method

ThomasFrans:
Since people were having problems when installing from crates.io, it would be nice to document it to prevent further problems.


ThomasFrans:
Thought it would be nice to add this to the documentation to try and help people when they install from crates.io, as there were some issues being created around this.

PR #1199, opened 2023-06-12
chore: improve release build options

### LINKED ISSUE #1192: Enable link time optimization for smaller builds
ncspot release builds are bigger than I would expect them to be. 20MB isn't gigantic, but it's bigger than I would expect, considering Alacritty for example only uses 6MB. Alacritty probably does a lot of dynamic linking so it might not be the best example, but 20MB still feels rather big.

A pretty good solution is to enable LTO during compilation. It drastically shrinks the binary from 21MB down to 14MB. Again, 21MB isn't that big, but most of the time ncspot is shipped as a binary which means the longer compile times would only be a drawback during the single compilation to create the package. It would only be beneficial for actual users themselves.

There is also "thin" lto which decreases the size to 17MB instead of 14MB. The linking time isn't that much shorter, so I don't think "thin" would make that much sense.

**Additional context**
I created a PR before that suggested changes to the build profiles and this was one of them. The PR was closed as it also removed the debug info in release builds, which was something that wasn't wanted. That's why I'm making this separate issue to discuss LTO, as I think it would only have a positive impact for release builds without any drawbacks.

I know I might be crazy to complain about 7MB, but when the program is already written in a well optimized, compiled language, it might as well go all the way and try to be as efficient as possible, especially considering the size decrease is completely free :smile:

ThomasFrans:
closes #1192

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,15 @@ members = [
     "xtask"
 ]
 
+[profile.release]
+lto = true
+codegen-units = 1
+
+[profile.optimized]
+inherits = "release"
+lto = false
+codegen-units = 16
+
 [dependencies]
 chrono = "0.4"
 clap = "4.2.7"


hrkfdn:
Merged, thanks!

juippis:
I noticed the buildtime for 0.13.4 has tripled compared to 0.13.3. If I revert this commit, it goes back to "normal".

Not a big deal as ncspot doesn't take too long to build anyway, just typing what I observed and if someone else sees the same :)

ThomasFrans:
I hoped that most people would get release binaries from their package managers so they wouldn't notice this change. Other people can either choose the longer compilation time or compile with the 'optimized' profile, which results in the old behavior.

PR #1201, opened 2023-06-14
fix: forward mouse events inside layout

ThomasFrans:
After refactoring layout, the if statements were removed in favor of a single match. This changed the control flow a bit, breaking mouse event propagation.

### DIFF
--- a/src/ui/layout.rs
+++ b/src/ui/layout.rs
@@ -260,7 +260,7 @@ impl Layout {
         }
 
         if is_left_right_event {
-            EventResult::Consumed(None)
+            EventResult::consumed()
         } else {
             result
         }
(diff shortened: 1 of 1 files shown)


ThomasFrans:
I'm very sorry that this made its way upstream. I have no clue how I didn't find this when testing everything. I only recently upgraded to a git build which did include this bug (it wasn't in v0.13.2), so that's how I noticed it.

I also removed the 'implicit' `EventResult::Consumed(None)` at the bottom and added it explicitly as it didn't make the code simpler to read at all (unlike I hoped it would). Also changed the `EventResult::Consumed(None)` to `EventResult::consumed()` to follow the same style that was already used before.

I hope this is the last bug from the refactor :crossed_fingers:

hrkfdn:
No problem, it happens! :) Thanks for fixing it.

PR #1205, opened 2023-06-18
refactor: remove unused `screenchange` variable

ThomasFrans:
The `screenchange` member variable on `Layout` didn't seem to be used for anything except for debugging. A comment in `on_layout` makes it seem like it wasn't meant to be used for debugging purposes but for actual functionality that wasn't implemented.

### DIFF
--- a/src/ui/layout.rs
+++ b/src/ui/layout.rs
@@ -11,7 +11,6 @@ use cursive::vec::Vec2;
 use cursive::view::{CannotFocus, IntoBoxedView, Selector};
 use cursive::views::EditView;
 use cursive::{Cursive, Printer};
-use log::debug;
 use unicode_width::UnicodeWidthStr;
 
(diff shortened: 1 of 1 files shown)


ThomasFrans:
Back with more insane space savings, a whole 1 byte this time :stuck_out_tongue_closed_eyes:

I was adding some documentation to `Layout` and wondered what this variable was for so looked it up, and except for the debug line, I don't see what it does. I think it might have been for future functionality that was never implemented. If it was for debugging purposes only, it makes sense, but a comment mentioned it was also used to let views know the active focus view was changed, something I can't find an implementation for. I could be wrong.

I just thought I'd remove it to clean up the code and avoid confusion about what this variable is for.

hrkfdn:
Indeed, looks like a leftover. Merged, thanks!

Issue #1198, opened 2023-06-09
Direct "save" on playing song


bashfulrobot:
**Is your feature request related to a problem? Please describe.**

This is not a problem per se. Currently, to save a song, it requires a few keystrokes. To save the currently playing song, you need to either `shift-o` and then arrow to `save` or go into the playlist, and arrow to the song and press `s`.

**Describe the solution you'd like**

Either or of these options would be great.

- a global single keystroke that will "save" the currently playing (vs selected) song.
- Or (and in some ways better), a cli flag to do the same. Then I could assign or script a global keyboard shortcut regardless of app selection. Or even if `ncspot` runs in the background in a `tmux` session. It would be trivial to call `ncspot` and then fire something like `notify-send` to display a confirmation.

**Describe alternatives you've considered**

- Just using the current options.
- maybe leaving ncspot open, and being "in" the playlist to use the arrow and `s` combo.

**Additional context**

- just showing selection vs currently playing.

![image](https://github.com/hrkfdn/ncspot/assets/10782881/bfc0b8e4-6173-47e1-b8e1-e8530494274c)

hrkfdn:
Implemented via #1220 by @hrdl-github

bashfulrobot:
:tada:  Thank you!

PR #1220, opened 2023-07-06
feat: add save current command

hrdl-github:
This touches upon #1198, but falls short of defining a key binding.


hrkfdn:
Merged, thank you!

PR #1238, opened 2023-07-21
Sort Artists Ignoring "The"

bentheklutz:
Ignore leading "The" when sorting albums and artists by artist name in the library.

e.g. "The Animals" now appears sorted under 'A' instead of under 'T' alongside other artists whose names begin with "The"


hrkfdn:
Thanks for the PR :heart: , I have made a very minor change. Merged :)

hrkfdn (approved):

PR #1258, opened 2023-08-20
Improve how we indicate that a command arg is optional

cyqsimon:
This is a minor PR with very minor improvements.


hrkfdn:
Looks good, thanks!

PR #1270, opened 2023-09-02
Feature: randomize queue

t-nil:
Hey,

so the motivation for this feature is that adding songs to play next breaks when you are on shuffle. Spotify solves this with a "second queue" which is immune to shuffle. However, going for less complexity, I have implemented a feature I knew from [foobar2000](https://www.foobar2000.org/): The ability to randomize the queue, that means, shuffling the elements of the backing queue Vec.

I tested the feature on my system in both debug and release mode. It seems to work.

### Design choices:
- **The playback is interrupted. If `ncspot` is playing when you execute the command, a `stop()` is executed and playback is restarted from the top of the queue as soon as randomizing is completed.** *Alternatives would have been to keep the current song playing, and either move it to the start, keep its position fixed or update the internal playback state, as after randomization the GUI was showing the title of the song which then occupied the same `id`. But stopping playback (and restarting after) seemed like the simplest, cleanest case.*
- **You can randomize the queue as often as you like.** *Another option would have been to save the old state and restore it, making randomization a toggle, but I assume that quickly gets messy when new songs are added in between etc.*

### TODO
- [ ] **Support DBus property / Fix `mpris`**
When I naively added the code in mpris.rs (which is now commented out), I got a lot of incomprehensive compiler errors, probably caused by complex macro expansion. One hypothesis of mine is that properties which have neither parameters nor return values are not supported. Since I have almost no experience with DBus I hope you can help me here @hrkfdn (or someone else ofc).
- [ ] **Implement randomization also for playlists.**
I don't know how much work that would be, and would probably be in favor of first accepting this PR and making a new one for that extra feature. But the odds of me finding time for that aren't too bad :)

@hrkfdn, would you be, in principal, willing to merge the feature described in this PR, and are there any changes needed for new commands I have overlooked? I just grep'd for `shuffle` and modified the same places (where it made sense).


wdoekes:
Randomize queue makes sense. I would prefer if you reconsidered the design decision of stopping/restarting play. A sudden stop in audio can be very noticable.

Putting the currently played song at pos 1 sounds like the logical thing to do to me.

t-nil:
Thanks for your input!

> I would prefer if you reconsidered the design decision of stopping/restarting play. A sudden stop in audio can be very noticable.

I agree, my gut tells me that's a very sensible design direction. I'd be down to implementing it.

Unfortunately, since my pull request is pretty old and I have not received any feedback from the maintainers, I would wait with committing further resources until that changes.

PR #1274, opened 2023-09-10
docs: update users.md

trimclain:
This will fix the description of the default keybinding <kbd>Ctrl</kbd>+<kbd>s</kbd>

Links:
[1] https://github.com/hrkfdn/ncspot/blob/main/src/commands.rs#L430
[2] https://github.com/hrkfdn/ncspot/blob/main/src/ui/queue.rs#L166


hrkfdn:
Merged, thank you!

PR #1288, opened 2023-09-27
style(clippy): enforce clippy `use_self` lint

### LINKED ISSUE #1175: Use clippy's `use_self` lint when it stabilizes
Currently the code has a lot of repeated names of structs (hundreds). I find one case when this is bad is when searching for the name of an item. If `Self` is used instead of repeating the name, searches across the code will show less noise from all of the repetitions.

As the `use_self` lint is still nightly-only, it doesn't make sense to add it right now. I just opened this as a sort of tracking issue for when it eventually stabilizes, as well as to discuss whether is would make sense to enable it at that point.

(I hope it's OK to create an issue as a reminder like this. There wasn't one yet. Otherwise it can be closed and be a future reference to show this was already discussed.)

ThomasFrans:
Clippy's `use_self` line ensures that `Self` is used instead of the real name whenever possible. This makes searching easier and cleans up the code a bit.

closes #1175

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -39,7 +39,7 @@ pub enum MoveAmount {
 
 impl Default for MoveAmount {
     fn default() -> Self {
-        MoveAmount::Integer(1)
+        Self::Integer(1)
     }
 }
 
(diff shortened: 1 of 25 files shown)


ThomasFrans:
This makes the MSRV 1.70. I hope that's OK.

hrkfdn (approved):
Thanks! LGTM! :)

Issue #1282, opened 2023-09-15
dump with tmux


Manix80:
**Describe the bug**
dump after closing

**To Reproduce**
kill tmux with ncspot inside

**Expected behavior**
do not disturb for the moment, this is just for information :)

**System (please complete the following information):**
 - OS: Linux archnico 6.5.3-zen1-1-zen
 - Terminal: kitty
 - Version: ncspot 0.13.4 (build in this repo)
 - Installed from: cargo

**Backtrace/Debug log**
~/.cache/ncspot/backtrace.log : 

```
   0: ncspot::panic::register_backtrace_panic_handler::{{closure}}
   1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/ec08a0337f3556212525dbf1d3b41e19bdf27621/library/alloc/src/boxed.rs:2021:9
   2: std::panicking::rust_panic_with_hook
             at /rustc/ec08a0337f3556212525dbf1d3b41e19bdf27621/library/std/src/panicking.rs:757:13
   3: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/ec08a0337f3556212525dbf1d3b41e19bdf27621/library/std/src/panicking.rs:631:13
   4: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/ec08a0337f3556212525dbf1d3b41e19bdf27621/library/std/src/sys_common/backtrace.rs:170:18
   5: rust_begin_unwind
             at /rustc/ec08a0337f3556212525dbf1d3b41e19bdf27621/library/std/src/panicking.rs:619:5
   6: core::panicking::panic_fmt
             at /rustc/ec08a0337f3556212525dbf1d3b41e19bdf27621/library/core/src/panicking.rs:72:14
   7: core::result::unwrap_failed
             at /rustc/ec08a0337f3556212525dbf1d3b41e19bdf27621/library/core/src/result.rs:1652:5
   8: core::ptr::drop_in_place<termion::raw::RawTerminal<std::io::buffered::bufwriter::BufWriter<std::fs::File>>>
   9: core::ptr::drop_in_place<core::cell::RefCell<termion::screen::AlternateScreen<termion::input::MouseTerminal<termion::raw::RawTerminal<std::io::buffered::bufwriter::BufWriter<std::fs::File>>>>>>
  10: core::ptr::drop_in_place<cursive::backends::termion::Backend>
[...]

PR #1299, opened 2023-10-04
fix: MPRIS loop `tokio::select!()` panic on exit

### LINKED ISSUE #1298: Backtrace file is created after every process exit
**Describe the bug**
Each time `ncspot` is quit, it creates a `backtrace.log` file. It should only create this file if it panics.

**To Reproduce**
1. Remove any existing `backtrace.log` file
2. Open and close `ncspot` normally
3. A new, empty `backtrace.log` file is created

**Expected behavior**
After a successful exit, no `backtrace.log` file should be created.

 - OS: Arch Linux
 - Terminal: Alacritty
 - Version: `master` branch
 - Installed from: /

**Additional context**
I added the code to generate the `backtrace.log` file a while ago, but I have no clue what is going wrong here. The exit code is `0` which leads me to believe the process exits successfully. The code to create the file is inside the panic handler function, which should only be executed when the program crashes in a normal way.

ThomasFrans:
This fixes a bug that would cause a panic when quiting the process normally. `tokio::select!()` was used to await a single branch, which is useless as it can be replaced by a normal await.

closes #1298

### DIFF
--- a/src/mpris.rs
+++ b/src/mpris.rs
@@ -513,13 +513,10 @@ impl MprisManager {
         let player_iface = player_iface_ref.get().await;
 
         loop {
-            tokio::select! {
-                Some(()) = rx.next() => {
-                    let ctx = player_iface_ref.signal_context();
-                    player_iface.playback_status_changed(ctx).await?;
-                    player_iface.metadata_changed(ctx).await?;
-                }
-            }
+            rx.next().await;
+            let ctx = player_iface_ref.signal_context();
+            player_iface.playback_status_changed(ctx).await?;
+            player_iface.metadata_changed(ctx).await?;
         }
     }
 


ThomasFrans:
I'm not 100% sure whether this is the correct way to solve it. I tested MPRIS with this patch and it still works fine. I also tested to make sure the backtrace file wasn't generated anymore. I'm still a bit unsure about how to correctly use `async` stuff... I think the `select!()` wasn't necessary as there was only a single `await`. I don't know why the `select!()` would cause a panic though.

ThomasFrans:
OMG I feel stupid... :rofl:
https://github.com/hrkfdn/ncspot/issues/1282#issuecomment-1745238164

hrkfdn:
Huh, you're right, not sure why there was a `select!` with a single branch.. :thinking: Thanks!

PR #1301, opened 2023-10-05
refactor: remove `lazy_static` crate

ThomasFrans:
The `lazy_static` crate was superseded by the `once_cell` crate which has been included in Rust's standard library since version `1.70`. Remove the `lazy_static` dependency and refactor all use cases to use `std::sync::OnceLock` instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1968,7 +1968,6 @@ dependencies = [
  "fern",
  "futures",
  "ioctl-rs",
- "lazy_static",
  "libc",
  "librespot-core",
  "librespot-playback",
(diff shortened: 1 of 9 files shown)


ThomasFrans:
This PR saves a whopping 96 bytes of space. You could fit an entire [BooBarBaz](https://github.com/hrkfdn/ncspot/pull/1301/files#diff-de6b9cfd1a256da7900d35e26aa7279bd713110e68e91b52616befb7397848dfL306-L308) in there!

hrkfdn:
Thanks!

PR #1303, opened 2023-10-07
fix: panic on `SIGHUP` by closed `/dev/tty`

### LINKED ISSUE #1282: dump with tmux

ThomasFrans:
This fixes a panic caused by the `SIGHUP` signal being sent to the process after its controlling terminal closes. Handling `SIGHUP` isn't necessary as we're not interested in whether the controlling terminal closed, only in the fact that we need to terminate. `SIGTERM` will always be sent as well so just react to that.

closes #1282

### DIFF
--- a/src/application.rs
+++ b/src/application.rs
@@ -7,7 +7,7 @@ use cursive::{Cursive, CursiveRunner};
 use log::{error, info, trace};
 
 #[cfg(unix)]
-use signal_hook::{consts::SIGHUP, consts::SIGTERM, iterator::Signals};
+use signal_hook::{consts::SIGTERM, iterator::Signals};
 
 use crate::command::Command;
 use crate::commands::CommandManager;
@@ -204,15 +204,14 @@ impl Application {
     /// Start the application and run the event loop.
     pub fn run(&mut self) -> Result<(), String> {
         #[cfg(unix)]
-        let mut signals =
-            Signals::new([SIGTERM, SIGHUP]).expect("could not register signal handler");
+        let mut signals = Signals::new([SIGTERM]).expect("could not register signal handler");
 
         // cursive event loop
         while self.cursive.is_running() {
             self.cursive.step();
             #[cfg(unix)]
             for signal in signals.pending() {
-                if signal == SIGTERM || signal == SIGHUP {
+                if signal == SIGTERM {
                     info!("Caught {}, cleaning up and closing", signal);
                     if let Some(data) = self.cursive.user_data::<UserData>().cloned() {
                         data.cmd.handle(&mut self.cursive, Command::Quit);


ThomasFrans:
This was an interesting one to find. I created a minimal version using just `cursive` and `signal-hook` which revealed that a minimal version shows the same behavior of panicking after receiving `SIGHUP`. I then removed the `SIGHUP` from the signals to handle and noticed the process still received `SIGTERM` when the terminal closed. I guess the kernel always sends signals in a specific order and makes sure to send `SIGTERM` BEFORE `/dev/tty` is closed, and `SIGHUP` AFTER `/dev/tty` is closed. It's a small thing but I didn't know it worked like this. Interesting bug :smile:

hrkfdn:
Does this also happen with the async handling from #1304? I'm wondering if there's some race condition due to duplicate signal handling and whether the way signals are handled (Cursive event loop vs. async task) makes a difference.

hrkfdn:
Does this also happen with the async handling from #1304? I'm wondering if there's some race condition due to duplicate signal handling and whether the way signals are handled (Cursive event loop vs. async task) makes a difference.

ThomasFrans:
You're right. I think the other PR solved the issue and this one isn't necessary. I just noticed that there is a bug in the current main branch because I mixed up these two PRs when I was working on them and one `SIGHUP` went missing in the `handle_signals()` function at `src/application.rs:64,20`.

PR #1304, opened 2023-10-07
feat: move to async POSIX signal handler

ThomasFrans:
Instead of trying to handle signals for every step of the `cursive` event loop, move the signal handling into its own asynchronous task and send callbacks to `cursive` when a signal arrives.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1985,6 +1985,7 @@ dependencies = [
  "serde_cbor",
  "serde_json",
  "signal-hook",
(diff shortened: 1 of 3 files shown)


ThomasFrans:
This is just a suggestion but I think it's a bit more efficient to use an asynchronous task to check for signals and it removes some clutter from the event loop.

hrkfdn:
Makes sense I think, thank you!

hrkfdn (approved):

PR #1305, opened 2023-10-08
fix: restore SIGHUP in signal handler

ThomasFrans:
Restore the SIGHUP signal handler which was accidentally removed in a067ab2.

PR #1310, opened 2023-10-11
Handle `SIGHUP` signals

### LINKED ISSUE #1309: ncspot does not close when terminal is closed
**Describe the bug**
The last commit a067ab2ae27e seems to break the way ncspot handles a SIGTERM. If I close the terminal ncspot continues to run and I have to kill it. Reverting this commit fixes this issue.

**To Reproduce**
1. Open ncspot in a terminal
2. Close the terminal
3. ncspot does not close

**Expected behavior**
ncspot should close as well, when the terminal is closed.

 - OS: Linux
 - Terminal: GNOME Terminal. kitty
 - Version: Master branch
 - Installed from: Compiled

hrkfdn:
While this probably reintroduces the crashes it should handle `SIGHUP` again which currently leads into the `unreachable!()` branch.

Fixes #1309

### DIFF
--- a/src/application.rs
+++ b/src/application.rs
@@ -61,7 +61,7 @@ async fn handle_signals(cursive_callback_sink: CbSink) {
     while let Some(signal) = signals.next().await {
         info!("Caught {}, cleaning up and closing", signal);
         match signal {
-            SIGTERM => {
+            SIGTERM | SIGHUP => {
                 cursive_callback_sink
                     .send(Box::new(|cursive| {
                         if let Some(data) = cursive.user_data::<UserData>().cloned() {


ThomasFrans:
This will not fix the issue. I've been looking for the issue and one of the first things I did was to add `SIGHUP` back to make sure the problem wasn't caused by that. The issue remains. I did some further digging in #1305 and added some findings as comments to that PR. If it's important to fix this for now, the smartest move might be to revert the commit.

Reverting won't fix the actual problem, it will just make it so `ncspot` panics like it did before the commit, which is better than consuming CPU in the background for sure :smile:

PR #1311, opened 2023-10-12
fix: disable `cursive_buffered_backend` tracing

ThomasFrans:
Disable the useless trace logs from the `cursive_buffered_backend` crate.


ThomasFrans:
A small change to the logging setup that is meant to clean up the debug logs quite a bit by removing all the useless `trace!()` logs from the `cursive_buffered_backend` crate.

hrkfdn:
Those are spammy indeed, thanks!

PR #1312, opened 2023-10-12
fix: prevent IPC socket cleanup panic

ThomasFrans:
Prevent a panic when the IPC socket isn't available when the user quits `ncspot`.


hrkfdn:
Merged, thanks!

PR #1319, opened 2023-10-24
chore: keep track of used toolchain version

ThomasFrans:
Keeping the used toolchain version in version control makes sure that when building older versions, the same toolchain is used every time. For example, when `ncspot 0.12.0` has always been compiled with Rust version `1.67.0` but the currently installed version is `1.73.0`, this file would cause Rustup to download the older version again.

### DIFF
--- a/.github/workflows/update-rust-toolchain.yml
+++ b/.github/workflows/update-rust-toolchain.yml
@@ -1,0 +1,21 @@
+name: update-rust-toolchain
+on:
+  workflow_dispatch: # allows manual triggering
+  schedule:
+    - cron: '0 9 * * wed' # runs at 9 am on every wednesday
+
+jobs:
+  update-rust-toolchain:
+    name: "Update Rust Toolchain (rust-toolchain.toml)"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: update rust toolchain
+        uses: a-kenji/update-rust-toolchain@v1.1
+        with:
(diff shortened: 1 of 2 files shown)


hrkfdn:
Hey, I think the idea is interesting, though it would be nice if there was a way to automatically update the version, i.e. via Dependabot, but apparently this is not supported yet: dependabot/dependabot-core/issues/1702

The zellij project has a GHA for this and uses the YAML file instead, it seems: https://github.com/zellij-org/zellij/blob/main/.github/workflows/update-rust-toolchain.yml

Maybe we should go for that approach as well to make sure it's always up to date?

ThomasFrans:
I think that's a valid approach. There is one problem that I could see which is that the first version after a toolchain update might not pass CI as that could require manually fixing new clippy issues. It's not the biggest problem, and you'd probably never checkout a 'toolchain-update-commit' anyway. I'll try to update this PR with that change.

ThomasFrans:
@hrkfdn 
I added a workflow. It's not tested yet. The workflow needs a secret to be set up in the repository, which is something you would need to do. The 'The workflow file' link below explains this in detail. Hope this is OK. After adding a secret, the workflow can be run manually (probably handy for testing) or will run every Wednesday at 9am. This might be the simplest for you to test it in tree with an outdated toolchain file and by manually triggering the workflow. It would create some noise commits but it will probably be the fastest. What do you think?

[The GitHub action](https://github.com/a-kenji/update-rust-toolchain)
[The workflow file](https://github.com/marketplace/actions/update-rust-toolchain-toml?version=v1.1#with-a-personal-authentication-token)

Issue #703, opened 2022-01-12
Cross platform media control


rashil2000:
**Is your feature request related to a problem? Please describe.**
I see that media control on Linux has been implemented via MPRIS. It would be great to have this functionality on all platforms.

**Describe the solution you'd like**
There's a crate - https://github.com/Sinono3/souvlaki - which abstracts away media control under a unified function. It not only supports Linux (via MPRIS), but also has full support for Windows and macOS.

**Describe alternatives you've considered**
None

**Additional context**
This is what a typical function looks like in souvlaki:

```rust
    controls
        .set_metadata(MediaMetadata {
            title: Some("When The Sun Hits"),
            album: Some("Souvlaki"),
            artist: Some("Slowdive"),
            duration: Some(Duration::from_secs_f64(4.0 * 60.0 + 50.0)),
            cover_url: Some("https://c.pxhere.com/photos/34/c1/souvlaki_authentic_greek_greek_food_mezes-497780.jpg!d"),
        })
        .unwrap();
```

See https://github.com/Sinono3/souvlaki/tree/master/examples for more examples.

jacksongoode:
This would be fantastic! 👍

itytophile:
I've investigated the problem. It seems to me that to have access to the Windows media controls we need to have a "handle to a window" (HWND). Unfortunately ncspot is a terminal app and I didn't find a way to get a handle that works with souvlaki. I've used [GetConsoleWindow](https://docs.microsoft.com/en-us/windows/console/getconsolewindow) (which is deprecated) without success.

rashil2000:
Hmm, doesn't the example code there seem to be using a console application?

itytophile:
To have a `MediaControl` instance (`controls` in your example) you have to provide a `PlatformConfig` that needs a HWND (https://docs.rs/souvlaki/latest/souvlaki/struct.MediaControls.html#method.new). On Windows it is required as you can see in [souvlaki source code](https://github.com/Sinono3/souvlaki/blob/30cedb9fb469ce53945689a7c3b5433a773d227a/src/platform/windows/mod.rs#L47).

rashil2000:
Ah, I see...

jacksongoode:
@itytophile Well getting it working for macOS/Linux systems would still be a win right?

itytophile:
@jacksongoode I already have media control on my linux desktop (GNOME) and I thought that the feature was available on macOS. I don't have an apple device to test that out.

[...]

PR #1327, opened 2023-10-29
Add cross platform media control keys support (#703)

ptixed:
## Describe your changes

Added new `media_control` feature making use of [souvlaki lib](https://github.com/Sinono3/souvlaki).  
Works on debian and windows 10. I will not be able to check on mac.  
I kept the old `mpris` feature but it is not compatible with `media_control`. Should it be removed?  
Code has been mostly adapted from current `mpris` implementation and https://github.com/aome510/spotify-player/.  

[#703](https://github.com/hrkfdn/ncspot/issues/703)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Special color for unavailable items
 - changelog with all the relevant user-facing changes to the project
+- Media control keys support for Windows and macOS
 
 ### Changed
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,22 @@
 version = 3
 
 [[package]]
+name = "ab_glyph"
+version = "0.2.22"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1061f3ff92c2f65800df1f12fc7b4ff44ee14783104187dd04dfee6f11b0fd2"
+dependencies = [
+ "ab_glyph_rasterizer",
+ "owned_ttf_parser",
+]
+
+[[package]]
+name = "ab_glyph_rasterizer"
+version = "0.1.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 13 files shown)


Bettehem:
If `media_control` supports the MPRIS features, I don't think there needs to be a separate `mpris` feature, so it can be removed.

ptixed:
So I looked at `mpris.rs` again and found some features that are not available in souvlaki:

* `supported_uri_schemes`
* `has_tracklist`
* `loop_status`
* metadata: `mpris:trackid`, `xesam:albumArtist`, `xesam:discNumber,` `xesam:trackNumber`, `xesam:url`, `xesam:userRating`
* `shuffle`
* `volume`
* `can_go_next`, `can_go_previous`, `can_play`, `can_pause`, `can_seek`, `can_control`

In souvlaki these mostly have hardcoded values or are nonexistent: https://github.com/Sinono3/souvlaki/blob/master/src/platform/mpris/zbus.rs

I only need to undo my last commit if we want to keep `mpris` as an option.

Bettehem:
Indeed, it seems like the Souvlaki MPRIS implementation is incomplete, so it won't work as a replacement for the current `mpris` feature as is.

I have two suggestions: Either implement MPRIS support in souvlaki properly, or use `media_control` only for Windows/macOS and `mpris` for others. I think the former is preferrable though, so that we won't have multiple libraries doing the same thing

ptixed:
@Bettehem I can't really work on souvlaki as I have no access to mac so I went with your second idea.

jacksongoode:
So excited to finally see this merged!

ptixed:
I've been using windows version for a week without issues. Anything else you need to accept the PR?

hrkfdn:
Sorry, currently moving to a different place, so super busy at the moment.. hope to have a look at this soon.

hrkfdn:
I would really like to merge this, but I think we should fix the souvlaki implementation first as @Bettehem said, so we can avoid having duplicate MPRIS logic in the codebase.

bogdan-calapod:
I'm with @veloek here - this would be a great QoL improvement! :D

PR #1338, opened 2023-11-20
fix: add leading slash to gitignore paths

ThomasFrans:
The leading slash forces Git to only look for the mentioned files from the root of the directory, not inside allowed subdirectories.


ThomasFrans:
Really small PR because I just learned the purpose of the leading slash in the gitignore file and remembered I didn't add them when I added the `xtask` stuff to generate assets.

Issue #1328, opened 2023-11-02
Changing volume internally does not change MPRIS volume


aubergino:
## Describe the bug
When I change the volume inside the ncspot application, whether by commands, scroll wheel, or keybinds, this does not update MPRIS' idea of ncspot's internal volume. As a matter of fact, the exact same one-sided communication happens with repeat and shuffle. For example...

## To Reproduce
1. Have a song playing so you can tell what's going on.
1. Use MPRIS (I use the Media Player Plasmoid from KDE Plasma) to set ncspot's volume to 100%.
2. Inside ncspot, set the volume (using any method you like) to 50%. The music gets quieter.
3. Use MPRIS to decrement the volume by 5%. It sets ncspot's volume to 95%.

## Expected behavior
In the above example, I expected ncspot's volume to be set to 45% after the final step.

## System
 - OS: Arch Linux
 - Terminal: Konsole
 - Version: 0.13.2-3
 - Installed from: Arch official repositories

## Screenshots
![](https://github.com/hrkfdn/ncspot/assets/113323010/8c3b2e42-775d-4bc1-a9da-09eb8ad62cad)
*Changed from inside ncspot, they don't match*

## Debug log
This log was massive, so for clarity I only included the part of the log that actually, y'know, relate to this bug report. I didn't exclude anything from where the log starts to where it ends, however; I only trimmed out the start and end.
[ncspot-debug-1.log](https://github.com/hrkfdn/ncspot/files/13241028/ncspot-debug-1.log)

### Status is passed on during startup
It seems that whenever you start ncspot, the state is loaded into MPRIS. For example, if I set repeat=R1, shuffle=on, volume=70% inside ncspot, MPRIS will have no idea. But if I quit ncspot then restart it, MPRIS is suddenly aware of all those things.
### `playerctl` works, so it might be partially KDE's fault...
I conducted some testing with `playerctl`, and it seems to me like half the fault here is with KDE. 
If I repeat the above methodology but replace the last step with running `playerctl volume -0.05`, *everything works fine*. 
Same goes for `playerctl shuffle toggle`.
`playerctl` doesn't have an absolute command for repeating, so I can't compare there.
My hypothesis, therefore, is that their Plasmoid does not pass on relative commands but appears to store an internal value for volume/repeat/shuffle and then use absolute commands like `playerctl volume 0.95`. Or `playerctl shuffle off` instead of `playerctl shuffle toggle`, depending on the internally stored value.
### ...but not completely
[...]

PR #1343, opened 2023-11-29
fix: allow any value to set MPRIS volume

ThomasFrans:
This fixes a small inconsistency between the MPRIS implementation and the specification. The specification allows any number when setting the volume, which have to be clamped to the effectively allowed range in the application.
https://specifications.freedesktop.org/mpris-spec/2.2/Player_Interface.html#Property:Volume

## Describe your changes
- Instead of only changing the number if it is in the range 0.0..0.1, always set it after clamping it to that range.

\/


ThomasFrans:
Found this while looking for a solution for #1328.

hrkfdn:
Thanks, looks reasonable!

Issue #1110, opened 2023-04-04
New MPRIS implementation has incorrect previous song behavior


eulerfan271:
**Describe the bug**
When using ``shift+,`` in ``ncspot`` in the middle of a song, the song starts over (this is the desired behavior) rather than going to the previous song.  For example, say song A plays, then song B plays, after 30 seconds in song B, I press ``shift+,``, then song B restarts.
With the new MPRIS implementation, interacting through MPRIS (specifically, with ``playerctl previous``) results in song A playing, unlike within the TUI.

This bug was not present with the old MPRIS implementation (rolling back only that commit fixes it), however I do not know enough about MPRIS to guess the problem beyond that.

**To Reproduce**
1. Open ncspot, play a song for 30 seconds
2. Call ``playerctl previous`` (or another MPRIS interface), nothing happens
3. Press ``shift+,`` in the TUI, song restarts from the beginning

**Expected behavior**
MPRIS previous behaves the same way as in the TUI and restarts the current song if further enough in.

eulerfan271:
Fantastic!

PR #1346, opened 2023-12-08
MPRIS: can_go_previous should always be true

eulerfan271:
## Describe your changes
Currently, the MPRIS previous does nothing when the song is the first in the queue, when it should restart the song (as is the case if one uses the `previous` keyboard shortcut in the TUI).  This can fixed by making the MPRIS function `can_go_previous` identically true, rather than checking if there's a previous song in the queue.

One might consider this is somewhat related to #1110?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - An unlikely crash when the UNIX IPC socket is removed before `ncspot` is closed
 - Guaranteed crash while quiting `ncspot` when using MPRIS
 - MPRIS volume not being updated when given numbers smaller than 0 or larger than 1
+- Allow previous track via MPRIS if first track in queue is playing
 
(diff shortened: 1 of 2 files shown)


hrkfdn:
Hey! Thanks for the PR. Should we at least check whether a track is loaded or similar, to disable the previous command if the queue is completely empty, for instance?

Also curious how other plays handle this. The specs aren't very specific about it..

eulerfan271:
Good point, thanks for the quick review!  I've updated it to be true only if something is in the queue; it now matches many of the other MPRIS functions (play, pause, etc).

As for other players, I think this behavior (calling "previous" on the first song in the queue repeats that song) is pretty common, and it's present in both ncspot (when using the commands in the TUI) and the official Spotify app.  In my opinion, it's also more useful/intuitive than the previous key doing nothing.

hrkfdn:
LGTM, let's get it in!

hrkfdn (approved):

PR #1360, opened 2024-01-01
style: deny `clippy::multiple_inherent_impls` lint

ThomasFrans:
## Describe your changes
Found there were two inherent `impl`s for `Worker`, for which there is apparently a Clippy lint.


ThomasFrans:
There was only one occurrence of this in the code, but it makes sense to check for it. This is such a weird feature in Rust that I personally don't see any advantage for. It does however cause some confusion as you could miss part of the functionality of a type when you are reading through the source code.

hrkfdn:
Merged, thanks! :)

PR #1361, opened 2024-01-01
chore: remove unnecessary asynchronous channel

ThomasFrans:
The channel does not need to be asynchronous as the receiving end is used in a blocking manner and the sending end of a regular channel never blocks.

## Describe your changes

- Remove the asynchronous futures `oneshot` channel in favor of a regular one


ThomasFrans:
I'm going through the `Worker` and `Spotify` structs code and documenting some stuff for future contributors (and myself to understand it better) and I'm just fixing some stuff I encounter along the way. I hope this change is OK. It seems weird to use an `async` channel when it is used to block on anyway. I tested the code by removing the cache and logging in again and everything seems to work like normal.

hrkfdn (approved):

Issue #1345, opened 2023-12-06
Crash on playback, Termux, Aarch64


WyntrHeart:
**Describe the bug**
Attempting to play any song instantly crashes ncspot without any error message.

**To Reproduce**
1. Install ncspot by doing 'pkg install ncspot'
2. Open ncspot
3. Attempt to play any song

**Expected behavior**
The song should play

**System (please complete the following information):**
 - OS: Android 9
 - Terminal: Termux
 - Version: v0.13.4
 - Installed from: Termux

**Backtrace/Debug log**
"https://github.com/hrkfdn/ncspot#debugging" doesn't exist, and I don't know where to obtain the relevant logs. If someone tells me how to obtain these logs I will provide them.
I can't do a backtrace because this issue is with a release build.

**Additional context**
I tried to work around the problem by building from source, but using either "cargo install ncspot" or "cargo build --release" from a source directory just hangs indefinitely for me on the second to last step.

galen8183:
Use the `-d` option to enable debug logging to a file.
See `ncspot --help` or `man ncspot` for more.

ThomasFrans:
> I can't do a backtrace because this issue is with a release build.

It would probably be useful regardless of the build. The backtrace also includes the reason why the program crashed, even though the backtrace itself won't be that helpful. It should be automatically generated in the  `ncspot` cache directory after a crash.

WyntrHeart:
So I hadn't touched ncspot at all since my previous post, and I just tried to start it to get the debug logs as instructed, but now it doesn't even open the player. The login screen works, but once I login it crashes immediately. I did find the backtrace and got a debug log, but the debug log may not show whatever caused the original issue since this new issue has appeared.
[ncspot-debug.log](https://github.com/hrkfdn/ncspot/files/13812884/ncspot-debug.log)
[backtrace.log](https://github.com/hrkfdn/ncspot/files/13812895/backtrace.log)

ThomasFrans:
Ah, I think I already see from the backtrace. This is something I added about a month or so ago. It's some code to look for a runtime directory for runtime files (like the IPC socket) that probably can't find any suitable runtime directory on Android. It used the user cache directory before from what I remember, but that shouldn't be used for runtime files so it was instead moved to directories like `/run/user/` or `/tmp/`. I don't know what the available directories are on Termux. I'll try to have a look and see what's going wrong.

WyntrHeart:
[...]

PR #1362, opened 2024-01-02
fix: crash on Termux due to missing runtime directory

ThomasFrans:
Instead of crashing on Termux, no IPC socket is created. This is a temporary solution until a suitable runtime directory for the Termux platform can be found.

## Describe your changes

- Make the IPC socket optional in case no suitable user runtime directory can be found on the platform.

#1345 


ThomasFrans:
I'll be honest: This isn't the best approach and it would be better to find a runtime directory but I'm not familiar at all with Termux so I have no clue how directories work on there. This should at least prevent a hard crash and will log an error instead.

hrkfdn:
It's fine I think, doesn't hurt to handle this gracefully

hrkfdn (approved):
Thanks! :)

PR #1363, opened 2024-01-04
fix: prevent 'called pause from invalid state' warning

ThomasFrans:
Calling `pause()` on the Librespot player while it is already paused generates a warning. Don't call pause when the player is already paused.

## Describe your changes

- Don't send `Pause` to the player when it's already paused during queue setup


ThomasFrans:
I think this is correct, but the code still seems a bit weird to me. I don't understand why it calls `pause()` on the `Spotify` struct if the song was loaded with `start_play` set to true. From what I understand, it would only need to check for the `PlaybackState::Stopped` case and stop the player if it matches. The others are all handled by loading with `playback_state == PlaybackState::Playing`. I don't fully understand how `PlaybackState::Default` works so I might be wrong here.

PR #1367, opened 2024-01-06
chore(actions): remove unsupported `pr-branch` input

ThomasFrans:
The latest version of the Rust toolchain update action supports a `pr-branch` input. It's not yet supported on `v1.1`.

## Describe your changes

- Remove the unsupported input from the action.


ThomasFrans:
I hadn't tested the action but I see that there is a small mistake. I used the action file as a guide to see what inputs were available, but forgot to use the one from `v1.1` which didn't have the `pr-branch` input yet. It also isn't correct, as I assumed it was used to input the master branch, which isn't the case.

hrkfdn:
Thanks for the PR! Yeah the action needs some maintenance upstream, `set-output` has been deprecated for a while.

PR #1368, opened 2024-01-06
fix(mpris): missing `PropertyChanged` signal for volume

### LINKED ISSUE #1328: Changing volume internally does not change MPRIS volume

ThomasFrans:
When the volume is changed from inside `ncspot`, also send a `PropertyChanged` signal.

## Describe your changes

- Send a DBus `PropertyChanged` signal every time the volume is changed inside `Spotify`
- Make the `update()` command on the `MprisManager` more general with different signals

closes #1328

### DIFF
--- a/src/mpris.rs
+++ b/src/mpris.rs
@@ -1,3 +1,4 @@
+use log::info;
 use std::collections::HashMap;
 use std::error::Error;
 use std::sync::Arc;
@@ -274,7 +275,7 @@ impl MprisPlayer {
         log::info!("set volume: {volume}");
         volume = volume.clamp(0.0, 1.0);
         let vol = (VOLUME_PERCENT as f64) * volume * 100.0;
-        self.spotify.set_volume(vol as u16);
+        self.spotify.set_volume(vol as u16, false);
         self.event.trigger();
     }
 
@@ -461,8 +462,19 @@ impl MprisPlayer {
     }
 }
 
+/// Commands to control the [MprisManager] worker thread.
+pub enum MprisCommand {
+    /// Notify about playback status and metadata updates.
+    NotifyPlaybackUpdate,
+    /// Notify about volume updates.
+    NotifyVolumeUpdate,
+}
+
+/// An MPRIS server that internally manager a thread which can be sent commands. This is internally
+/// shared and cloning it will yield a reference to the same server.
+#[derive(Clone)]
 pub struct MprisManager {
-    tx: mpsc::UnboundedSender<()>,
+    tx: mpsc::UnboundedSender<MprisCommand>,
 }
 
 impl MprisManager {
@@ -480,7 +492,7 @@ impl MprisManager {
             spotify,
         };
 
-        let (tx, rx) = mpsc::unbounded_channel::<()>();
+        let (tx, rx) = mpsc::unbounded_channel::<MprisCommand>();
 
         ASYNC_RUNTIME.get().unwrap().spawn(async {
             let result = Self::serve(UnboundedReceiverStream::new(rx), root, player).await;
@@ -493,7 +505,7 @@ impl MprisManager {
     }
 
     async fn serve(
-        mut rx: UnboundedReceiverStream<()>,
+        mut rx: UnboundedReceiverStream<MprisCommand>,
         root: MprisRoot,
         player: MprisPlayer,
     ) -> Result<(), Box<dyn Error + Sync + Send>> {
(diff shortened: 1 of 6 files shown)


ThomasFrans:
Two little gotchas with this PR:

1. The signal is sent twice when changing volume over MPRIS. I didn't think it was worth it to duplicate the logic in `Spotify::set_volume()` to solve this as that will probably cause more trouble than it will solve in the long run. I don't see a problem with sending it twice, and there isn't a way to disable it from what I can tell. It this needs to be solved, let me know :)
2. The implementation is a bit 'un-rustlike' because `Spotify` depends on `MprisManager` now and `MprisManager` on `Spotify`, so the `mpris` value in `Spotify` is optional for this reason, to set it afterwards. I don't know of a better way to do this for now.

src/mpris.rs
```diff
@@ -462,8 +463,19 @@ impl MprisPlayer {
     }
 }
 
+/// Commands to control the [MprisManager] worker thread.
+pub enum MprisCommand {
+    /// Notify about playback status and metadata updates.
+    PlaybackUpdated,
+    /// Notify about volume updates.
+    VolumeUpdated,
```
hrkfdn, line 471: Nit: These read more like event names than commands, WDYT?

ThomasFrans: I completely agree. I should have worked a bit more on this PR before creating it as I don't feel 100% happy with it to be honest. It feels rushed. I'm going to turn it into a draft and have another look at it after the 25th (unless you want to fix the issue earlier of course). I'm currently a bit busy with my studies 🫤

hrkfdn: No worries, as always, appreciate your contributions and take your time. Good luck with your studies!

ThomasFrans:
I changed the names of the commands and prevented the duplicate message when updating the volume through the MPRIS interface. I will work on the events vs state in another PR to not make this one too big. Before touching the configuration stuff I would _really_ like to add tests to that module because I'm scared to break people's configs if I start changing related types.

Let me know if this PR is ok or you want to solve it in a different way :slightly_smiling_face:

hrkfdn (approved):
Thank you, LGTM. I was wondering if there's a way to reduce some of the `#[cfg]` conditionals, but I don't see how. Let's get it in! :)

ThomasFrans:
> I was wondering if there's a way to reduce some of the `#[cfg]` conditionals, but I don't see how. Let's get it in! :)

You don't want to know how much I've thought about this as well. It's hard to get right because there are too many combinations and also because any code that isn't used causes clippy warnings which in turn cause CI to fail. The only solution I can see is to use libraries which makes the conditional stuff "someone else's problem" 😄

PR #1381, opened 2024-01-28
docs: small overall documentation improvements

ThomasFrans:
General improvements to documentation and code clarity.

## Describe your changes
- Add documentation comments to various items
- Change web API return types from bool/Option to Result
- Create helper functions with descriptive names instead of comments
- Remove redundant/confusing types
- Fix some documentation comments as instructed by `cargo doc`
- Rename variables to clear names

### DIFF
--- a/src/library.rs
+++ b/src/library.rs
@@ -21,11 +21,20 @@ use crate::model::show::Show;
 use crate::model::track::Track;
 use crate::spotify::Spotify;
 
+/// Cached tracks database filename.
 const CACHE_TRACKS: &str = "tracks.db";
+
+/// Cached albums database filename.
 const CACHE_ALBUMS: &str = "albums.db";
+
+/// Cached artists database filename.
 const CACHE_ARTISTS: &str = "artists.db";
+
+/// Cached playlists database filename.
 const CACHE_PLAYLISTS: &str = "playlists.db";
 
+/// The user library with all their saved tracks, albums, playlists... High level interface to the
+/// Spotify API used to manage items in the user library.
 #[derive(Clone)]
 pub struct Library {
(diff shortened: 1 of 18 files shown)


ThomasFrans:
I'm adding documentation because I like to work on the project with `cargo doc --document-private-items` open to get a general overview of a type's functionality. It also helps a lot with language server hover to quickly get information about functions. I think it's a help for new contributors as well to more easily get an idea of what does what.

I changed some code (most notably the web API return types) when it made sense instead of adding comments/documentation. For example, it makes more sense to return `Result<(), ()>` instead of `bool` because the former makes clear that the function can fail and how its status is returned, while the latter needs documentation to explain what `true` or `false` mean. I don't think it has any performance impact as the compiler probably optimizes the abstractions away.

src/spotify.rs
```diff
@@ -1,37 +1,37 @@
...
 use crate::events::{Event, EventManager};
 use crate::model::playable::Playable;
 use crate::spotify_api::WebApi;
 use crate::spotify_worker::{Worker, WorkerCommand};
 
+/// One percent of the maximum supported [Player] volume, used when setting the volume to a certain
+/// percent.
 pub const VOLUME_PERCENT: u16 = ((u16::max_value() as f64) * 1.0 / 100.0) as u16;
 
+/// Events sent by the
```
ThomasFrans, line 34: .

src/library.rs
```diff
@@ -813,12 +828,18 @@ impl Library {
         }
     }
 
+    ///
```
ThomasFrans, line 831: .

hrkfdn (approved):
:pray:

PR #1382, opened 2024-01-28
refactor: general small refactors to simplify code

ThomasFrans:
Some small general refactoring to improve code readability and adhere to Rust conventions.

## Describe your changes
- Remove `expect` in favor of `unwrap` when the `Result`'s error variant contains the info in the `expect` anyway (eg. when locking things). The line number/context are given by the backtrace.
- Remove over-specification of types (`&T` instead of `&RWReadLockGuard`)
- Put reused values into constants
- `try_from` instead of manual function
- Change `if let Some(()) = ...` to `if T.is_some()`

### DIFF
--- a/src/spotify.rs
+++ b/src/spotify.rs
@@ -1,7 +1,8 @@
-use std::env;
+use std::error::Error;
 use std::str::FromStr;
 use std::sync::{Arc, RwLock};
 use std::time::{Duration, SystemTime};
+use std::{env, fmt};
 
 use futures::channel::oneshot;
 use librespot_core::authentication::Credentials;
@@ -56,8 +57,6 @@ pub struct Spotify {
     since: Arc<RwLock<Option<SystemTime>>>,
     /// Channel to send commands to the worker thread.
     channel: Arc<RwLock<Option<mpsc::UnboundedSender<WorkerCommand>>>>,
-    /// The username of the logged in user.
-    user: Option<String>,
 }
 
 impl Spotify {
@@ -71,19 +70,18 @@ impl Spotify {
             elapsed: Arc::new(RwLock::new(None)),
             since: Arc::new(RwLock::new(None)),
             channel: Arc::new(RwLock::new(None)),
-            user: None,
         };
 
         let (user_tx, user_rx) = oneshot::channel();
         spotify.start_worker(Some(user_tx));
-        spotify.user = ASYNC_RUNTIME.get().unwrap().block_on(user_rx).ok();
+        let user = ASYNC_RUNTIME.get().unwrap().block_on(user_rx).ok();
         let volume = cfg.state().volume;
         spotify.set_volume(volume);
 
         spotify.api.set_worker_channel(spotify.channel.clone());
         spotify.api.update_token();
 
-        spotify.api.set_user(spotify.user.clone());
+        spotify.api.set_user(user);
 
         spotify
     }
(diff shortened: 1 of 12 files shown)


ThomasFrans:
These are some things I encountered while adding documentation in #1381. The only one that I'm not sure about is the removal of `expect()` in favor of `unwrap()`. The reasoning as explained in the PR message is that `unwrap()` provides the same information as `expect()` but looks a bit cleaner. The text in the `expect()` doesn't really add that much valuable information in the case of unlocking poisoned synchronization types.

ThomasFrans:
Quickly changed the `UriType` parsing from `From<&str>` to `FromStr` as it's parsing the string.

hrkfdn:
Thanks! Back then I thought the tiny extra context that `expect()` brings is beneficial, but I guess it's sort of pointless if you have a full stack trace.

ThomasFrans:
There is the stack trace to point out the location and there is the general message of the error that says what went wrong. I think one of the next Rust releases is going to remove a bunch of debug information by default by stripping every binary, whether it's in the profile or not. You could always disable the stripping again if that happens. I quickly checked what another project like Alacritty does for locking and sending to channels and it seems like they also use `unwrap()`. Also checked Zellij and they use `expect()` for receiving from channels but `unwrap()` for locking. In the receive, they always use the same message so it doesn't give any context either.

src/spotify.rs
```diff
@@ -71,19 +70,18 @@ impl Spotify {
...
-        spotify.user = ASYNC_RUNTIME.get().unwrap().block_on(user_rx).ok();
+        let user = ASYNC_RUNTIME.get().unwrap().block_on(user_rx).ok();
         let volume = cfg.state().volume;
         spotify.set_volume(volume);
 
         spotify.api.set_worker_channel(spotify.channel.clone());
         spotify.api.update_token();
 
-        spotify.api.set_user(spotify.user.clone());
+        spotify.api.set_user(user);
```
hrkfdn, line 84: Probably not a problem, but this will set the user later than before. A quick check didn't give me anything, so I think this is fine.

src/spotify.rs
```diff
@@ -434,23 +410,59 @@ pub enum UriType {
...
+        assert!(matches!(uri_type, Err(UriParseError)));
+    }
+
+    #[test]
+
+    fn parse_playlist_uri() {
+        let uri_type = "spotify:playlist:37i9dQZF1DX36Xw4IJIVKA".parse();
+        assert!(matches!(uri_type, Ok(UriType::Playlist)));
+    }
+}
```
hrkfdn, line 468: Thanks :pray:

hrkfdn (approved):
Thanks, let's get it in! :)

src/spotify.rs
ThomasFrans: I don't fully understand. I thought nothing would change because `Spotify::user` was only used in the `Spotify::new()` method?

hrkfdn: Huh, yeah, I think I may have misread that. Sorry for the noise!

ThomasFrans: No I understand. I grepped the project so I knew it wasn't used anywhere else. I can imagine you don't know every line off the top of your head :smile:

(PS In the next days I'll probably try to solve/continue work on some of the bugs that have shown up in the last couple of months, if that's ok :slightly_smiling_face:)

PR #1387, opened 2024-01-31
docs(worker): comment purpose of UI refresh task

ThomasFrans:
The purpose of the manual ui refresh task isn't immediately clear. It was explained in the commit message of the commit that introduced it.

## Describe your changes

Add a comment that explains the purpose of the `refresh_ui` task in the worker thread.

### DIFF
--- a/src/spotify_worker.rs
+++ b/src/spotify_worker.rs
@@ -186,6 +186,7 @@ impl Worker {
                     },
                     _ => {}
                 },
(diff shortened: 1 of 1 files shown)


ThomasFrans:
Stupidly small PR but I had to 'Git blame' to find what this was used for. Thought I'd add a comment in case others read this and didn't understand either. I assumed the player events were used to update the status bar.

PR #1388, opened 2024-02-04
ci: Run tests in CI

hrkfdn:
## Describe your changes

Thanks to contributors test coverage is becoming a thing, thus it makes sense to run tests in CI workflows.

Additionally, switch back to `cargo build` instead of `cargo check`, as the tests need compilation anyway.

- [X] Documentation was updated (i.e. due to changes in keybindings, commands, etc.)
- [X] Changelog was updated with relevant user-facing changes (eg. not dependency updates,

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -55,7 +55,7 @@ jobs:
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
         with:
-          target: ${{ matrix.target }}
+          targets: ${{ matrix.target }}
           toolchain: stable
       - name: Install macOS dependencies
         if: matrix.os == 'macos-latest'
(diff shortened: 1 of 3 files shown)


ThomasFrans:
I'm not 100% sure about this but doesn't the `test` build use a different build profile? It does inherit the `dev` profile, but I don't know whether `cargo` is smart enough to reuse the same build artifacts in practice...

hrkfdn:
Hmm, I tested on my machine and it only linked the binary and didn't need
to rebuild the crates, but maybe there's a cleaner way.

I also thought of simply running `cargo test` only, as that would check the
build by itself.

ThomasFrans:
Yes I was wrong. I should have tested first. I just assumed that cargo treated both differently.

PR #1395, opened 2024-02-09
fix(cargo): prevent crate filename collision

ThomasFrans:
`cargo` has a bug that prevents correct handling of similarly named crates in the same package. Rename the internal library crate to prevent error messages. https://github.com/rust-lang/cargo/issues/10368

## Describe your changes

Rename the library crate in the `ncspot` package. Instead of the implicit name `ncspot`, give it a name in the `cargo` manifest.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,9 @@ inherits = "release"
 lto = false
 codegen-units = 16
 
+[lib]
+name = "ncspot_common"
+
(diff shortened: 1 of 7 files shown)


ThomasFrans:
This solves a minor issue when generating documentation for the binary crate. Without this fix, it throws a warning like in the following image.

image

Issue #1358, opened 2024-01-01
ncspot hangs after a period of inactivity


mcookcomcast:
**Describe the bug**

I am running ncspot 1.0.0 (312f9ff) on a raspberry pi

`Linux raspberrypi 5.10.63-v7l+ #1496 SMP Wed Dec 1 15:58:56 GMT 2021 armv7l GNU/Linux`

I control ncspot via mpris. 

An example command I send to play a track is:
`dbus-send --print-reply --dest=org.mpris.MediaPlayer2.ncspot /org/mpris/MediaPlayer2 org.mpris.MediaPlayer2.Player.OpenUri "string:https://open.spotify.com/track/3MZn1thdvB7QCYXY1AiTjG"`

When I start ncspot, it works fine. If I leave it running for some period of time*, it becomes unresponsive to requests to play tracks. The UI still allows navigation. I'm able to move from tab to tab and I'm able to launch the queue view and go back to the library view. However, if I attempt to run the above dbus-send command, it times out with an error:
```
pi@raspberrypi:~ $ dbus-send --print-reply --dest=org.mpris.MediaPlayer2.ncspot /org/mpris/MediaPlayer2 org.mpris.MediaPlayer2.Player.OpenUri "string:https://open.spotify.com/track/3MZn1thdvB7QCYXY1AiTjG"
Error org.freedesktop.DBus.Error.NoReply: Did not receive a reply. Possible causes include: the remote application did not send a reply, the message bus security policy blocked the reply, the reply timeout expired, or the network connection was broken.
```

The UI is also unable to play. If I hit ENTER on a track in the library, it does not play the track.

I have also tried to control it via nc:

`pi@raspberrypi:~ $ nc -U /run/user/1000/ncspot/ncspot.sock
[...]

PR #1396, opened 2024-02-10
attempt to fix https://github.com/hrkfdn/ncspot/issues/1358

eprst:
## Describe your changes

mpris runs inside `tokio::runtime.spawn`. At some point it hits `metadata_changed->WebApi::track->api_with_retry`, which gets 401 response and tries to update the token.

This goes into `update_token`, which at a certain point calls *blocking* `token_rx.recv()`. Tokio runtime is not happy about that, as it uses the same thread for the `run_loop` in the worker. That's why `tokio::select!` doesn't work, we're essentially deadlocking, so the worker never pick up `RequestToken` command and `update_token` block forever.

Here I'm changing `update_token` to wrap `recv` in `spawn_blocking`, which makes `update_token` return a `JoinHandle`, on which the caller has to await. This doesn't work nicely in case of mpris though, as it is not an async function, and I don't know how to make it async as it goes through the dbus `metadata_changed` interface. My best effort here is to do `and_then`, which seems to work, but I'm not confident that's the right approach.

I didn't do extensive testing, but so far seems to work.

https://github.com/hrkfdn/ncspot/issues/1358

### DIFF
--- a/src/spotify.rs
+++ b/src/spotify.rs
@@ -83,7 +83,11 @@ impl Spotify {
         spotify.set_volume(volume);
 
         spotify.api.set_worker_channel(spotify.channel.clone());
-        ASYNC_RUNTIME.get().unwrap().block_on(spotify.api.update_token().unwrap()).ok();
+        ASYNC_RUNTIME
+            .get()
+            .unwrap()
+            .block_on(spotify.api.update_token().unwrap())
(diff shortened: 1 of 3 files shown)


hrkfdn:
Thanks! The change makes sense, let's get it in for further evaluation

PR #1406, opened 2024-03-02
fix a panic when token update fails

eprst:
## Describe your changes
sometimes we get an `Err` from `token_rx.recv()`, shouldn't panic in such cases

didn't create an issue


hrkfdn:
Thanks, merged.

hrkfdn (approved):

PR #1412, opened 2024-03-09
Bump rspotify to 0.13.0

### LINKED ISSUE #1384: Requests leak out of HTTP proxy
## Description

When running ncspot with `http_proxy` environment variable set, most requests are proxied correctly, but some requests to `api.spotify.com` (which is `CNAME`-ed to `edge-web.dual-gslb.spotify.com`) are not. This is problematic if the network environment requires the use of a proxy for external network access.

Specifically, features like search and favourites are not accessible, although if a track is already selected it can continue to play (because the track streaming API calls are correctly proxied).

## Steps to reproduce

1. On a host machine, install privoxy (or another proxy server of your choice).
2. Configure privoxy to listen on all interfaces (in `/etc/privoxy/config`, set the value of `listen-address` to `:8118`).
3. Start privoxy.
4. Open up a quiet VM, ideally Linux. Make sure VM can access host's privoxy instance (check host's firewall?).
5. On the VM, install ncspot and wireshark; start wireshark.
6. In wireshark, set the following filter: `tcp && ip.dst != <VM_IP>`.
7. In wireshark, enable "View > Name Resolution > Resolve Network Addresses".
8. Start ncspot like so: `http_proxy=http://<HOST_IP>:8118 ncspot` (or whatever proxy address you are using).
9. Observe in wireshark's captures that while some requests are sent to host (these are correctly proxied), some requests are sent directly to `edge-web.dual-gslb.spotify.com`.

If you want to see what happens when a direct connection to `api.spotify.com` is not possible, simply remove the default route in your VM: `sudo ip route del default`. If you are using NetworkManager, you may have to do this by setting "IPv4 > Use only for resources on this connection" for the active network profile, then reconnecting the network.

## Expected behaviour

All requests of ncspot are proxied.

## Screenshots

wireshark capture

## System

 - OS: Endeavour OS, 6.7.2-arch1-1 x86_64
 - Terminal: Alacritty
 - Version: 1.0.0
 - Installed from: cargo

## Debug log

For this log, I have blocked direct connection to `api.spotify.com` so that these requests can error.

[ncspot.log](https://github.com/hrkfdn/ncspot/files/14080635/ncspot.log)

## Additional context

I have also tried setting `https_proxy` and `ALL_PROXY`, but it seems like the only one that has any effect is `http_proxy`.

cyqsimon:
Fixes #1384

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+- All requests are correctly proxied when the relevant environment variables are set
+
 ## [1.1.0] - 2024-03-05
 
 ### Added
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -428,6 +428,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9d297deb1925b89f2ccc13d7635fa0714f12c87adce1c75356b39ca9b7178567"
 
 [[package]]
+name = "base64"
+version = "0.22.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


cyqsimon:
Clippy complaint is irrelevant so I'll ignore them for this PR.

hrkfdn (approved):

PR #1420, opened 2024-03-21
added support for ap-port conf

### LINKED ISSUE #1404: add support for --ap-port option
This would solve issues with restrictive firewalls which block connections in the `4070` port but allow it on `443`, which explains #90 and #1229

`librespot` supports `--ap-port 443` option, please make it available as an option to `ncspot` as well

PS:
I tested simply adding a line `session_config.ap_port = Some(443);` in function 
https://github.com/hrkfdn/ncspot/blob/c0b78d3a2a1ae7046e296ad8ef6d727ab8d8c7be/src/spotify.rs#L131-L141
and it works on my environment. Adding proper support would require adding a config/command line option though.

gilcu3:
## Describe your changes
Added a config option `ap-port` to fix the port used to connect to `spotify` and pass it to `librespot`. Fixes issues with restrictive firewalls, after setting `ap-port = 443` in config file.
Fixes #1404 

### DIFF
--- a/src/spotify.rs
+++ b/src/spotify.rs
@@ -128,7 +128,7 @@ impl Spotify {
     }
 
     /// Generate the librespot [SessionConfig] used when creating a [Session].
-    pub fn session_config() -> SessionConfig {
+    pub fn session_config(cfg: &config::Config) -> SessionConfig {
         let mut session_config = SessionConfig::default();
         match env::var("http_proxy") {
             Ok(proxy) => {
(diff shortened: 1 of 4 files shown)


src/spotify.rs
```diff
@@ -137,12 +137,19 @@ impl Spotify {
             }
             Err(_) => debug!("No HTTP proxy set"),
         }
+        let ap_port = cfg.values().ap_port.unwrap_or(0);
+        if ap_port != 0 {
+            session_config.ap_port = Some(ap_port);
+        }
```
hrkfdn, line 143: Do we need this intermediate `0` value? Could we also do it like this?
```suggestion
        session_config.ap_port = cfg.values().ap_port;
```

gilcu3: I was not sure if that would behave in the correct way (no parameter is set) when the `ap-port` option is missing in the config file. If that's the case, please feel free to apply the change, I wont be able in a couple of days.

hrkfdn: By default librespot sets `ap_port` to `None`: https://github.com/librespot-org/librespot/blob/4d35c5ffe222c839d1532d8afac78fe4c37043b3/core/src/config.rs#L33

However, if we wanted to be absolutely sure we could also use an `if let` expression here.

Issue #1159, opened 2023-05-10
Playlists, podcasts and browse not loading


ThomasFrans:
**Describe the bug**
Playlists, podcasts and the browse tab aren't loading anymore. I quickly bisected since v0.13.0 to main and it appears that 80da5a8 introduced the bug. I have tried to revert only the commit that seems to break it and indeed playlists load again.

**To Reproduce**
1. Checkout 80da5a8
2. Go to the playlists tab (could be that you have to clear cache, I don't)
3. Nothing loads

**Screenshots**
![image](https://github.com/hrkfdn/ncspot/assets/48214567/23a6651f-54a3-41d8-a416-b6441f866719)

**System (please complete the following information):**
 - OS: Arch Linux (latest updates)
 - Terminal: Alacritty
 - Version: 80da5a8 and further
 - Installed from: source

ThomasFrans:
It also seems like something went wrong with the commit before that one (22badc9) as it doesn't compile.

hrkfdn:
Thanks. Can reproduce it, I've reverted the switch to native TLS for now.

jirutka:
Can you provide more details, like a log…?

jirutka:
`ncspot` is still both dynamically linked with the system TLS library and statically linked with Rustls, which is clearly wrong. It should use one or another (preferably the system library, maybe with an option to switch to bundled Rustls), not both.

jirutka:
ncspot 0.13.4:

```
[2024-03-31][18:59:25] [ureq::stream] [DEBUG] connecting to api.spotify.com:443 at [2600:1901:1:c36::]:443
[2024-03-31][18:59:25] [ncspot::spotify_api] [DEBUG] http error: Transport(Transport { kind: UnknownScheme, message: Some("cannot make HTTPS request because no TLS backend is configured"), url: Some(Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("api.spotify.com")), port: None, path: "/v1/me/", query: None, fragment: None }), source: None })
```

PR #1431, opened 2024-04-02
Use native TLS only, don't mix it with rustls (2nd try)

jirutka:
reqwest uses native TLS, but rspotify with ureq currently uses rustls, i.e. ncspot bundles rustls and is linked with the system TLS library at the same time.

This has already been fixed in 80da5a8, but this commit has been reverted in aeff120 due to a bug in rspotify that was fixed in 0.13.1 (ramsayleung/rspotify#471).

This time I properly tested that it works.

Related to #1159 #1153


hrkfdn:
Thanks for taking care!

PR #1439, opened 2024-04-09
Clippy: warn `enum_glob_use`

cyqsimon:
## Describe your changes

Warn `enum_glob_use`. This mostly has to do with my own code.

The rationale behind this is best explained by [this video](https://youtu.be/8j_FbjiowvE?t=97).

No user facing changes.

### DIFF
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::enum_glob_use)]
+
 use cursive::theme::BaseColor::*;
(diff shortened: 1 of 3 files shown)


src/theme.rs
cyqsimon, line 1: It's reasonable to allow it in this file because we are not using the variants in match blocks.

hrkfdn:
That's wild, I wasn't aware of this. Thanks!

hrkfdn (approved):

PR #1451, opened 2024-05-07
cfg fix for linux

aragonnetje6:
## Describe your changes
changed cfg!(linux) to cfg!(target_os = "linux") since the former is not actually valid, but produces no bug. this wasn't causing any problem that i know of, but if the code is in here it should probably do what it says, or be removed

no issue created, just saw a warning when installing from master (though this warning is not showing up again for some reason)

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -103,7 +103,7 @@ pub fn user_runtime_directory() -> Option<PathBuf> {
 
     if let Some(xdg_runtime_directory) = xdg_runtime_directory() {
         Some(xdg_runtime_directory.join("ncspot"))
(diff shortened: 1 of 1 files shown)


ThomasFrans:
Oops, I should have read the documentation of the `cfg!()` macro better :grimacing:
Nice find. I wonder why this every worked for me as I assume `cfg!(linux)` just doesn't do anything, but if I remember correctly the code did work on FreeBSD for example. I installed it in a VM specifically to test this code. Maybe I'm just misremembering :thinking:

hrkfdn (approved):
Good catch indeed!

PR #1454, opened 2024-05-14
style(editorconfig): add EditorConfig file for consistent style

ThomasFrans:
Add an EditorConfig file to ensure all editors know how to edit files correctly. Apply conservative settings based on the existing files. Rust options are based on the defaults of `rustfmt`.

Wondering whether the repository should include one of those `.vscode/settings.json` files to recommend the EditorConfig extension as Visual Studio Code doesn't come with support for EditorConfig by default. I'm not a fan of adding editor-specific files to a repository but I guess it would benefit the project if files (other than Rust files which are formatted anyway) have a consistent style.

## Describe your changes
- Add an `.editorconfig` file with options based on existing files

\/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
+charset = utf-8
+insert_final_newline = false
+
+[*.rs]
+max_line_length = 100
+
+[*.md]
+max_line_length = 100


hrkfdn (approved):
Thanks!

hrkfdn:
> Wondering whether the repository should include one of those .vscode/settings.json files to recommend the EditorConfig extension as Visual Studio Code doesn't come with support for EditorConfig by default.

No strong feelings here as VSCode is pretty common, though I agree with your sentiment about including editor-specific files. If it helps with your setup that's fine, though.

ThomasFrans:
I'll add it as it's the most common editor used today. I don't use it myself, but I do know that it does not have support for EditorConfig, although it's pretty good at figuring out the indentation style of a file on its own, which is enabled by default.

PR #1477, opened 2024-07-03
docs: Rename `ap-port` option as `ap_port`

Jiogo18:
#1420 introduced the option `ac_port` to specify the port used by librespot.
This option works great, except when following the documentation and example only.

- The option is named `ap-port` in librespot.
- The option is named `ap_port` in [ConfigValues](https://github.com/hrkfdn/ncspot/blob/9624c03264bb4038138f03374c4bfb70db7bc850/src/config.rs#L105).
- However, the option is named `ac-port` in the documentation and in #1420, which doesn't work, as described below.

In ncspot/config.toml:
- ❌ `ap-port = 443` does nothing, librespot uses port 4070
- ✅ `ap_port = 443` librespot uses HTTPS

## Describe your changes
Rename `ap-port` option by `ap_port` in the documentation.

## Before merge
Should the changelog be updated too?
Changelog of 1.1.1 contains `ap-port`, which is misleading.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [1.1.1] - 2024-05-10
 
 ### Added
-- `ap-port` configuration variable to choose what Spotify access point port to use
+- `ap_port` configuration variable to choose what Spotify access point port to use
 - Instructions for installation with `snap`
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


ThomasFrans:
I'd update the changelog as well. When I added it, I based it on the keepachangelog format which states that changes to earlier parts are encouraged if they improve the changelog. It's nice if users read the changelog that it kind of serves as "brief documentation" for new features 🙂

hrkfdn:
Thank you, merged :)

PR #1493, opened 2024-08-03
Update cursive to 0.21.0

gyscos:
## Describe your changes

This updates cursive to 0.21.0. This release includes a buffered backend internally, so there is no need for a separate `cursive_buffered_backend` crate.

The main change is that `View` now require `Send + Sync`, so generic views should forward these requirements.
Note that the actual views here were already `Send + Sync`, and no logic change was needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ main ]
 
+env:
+  TERM: xterm
+
 jobs:
   # Build the Rust code.
   build:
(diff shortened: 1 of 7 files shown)


gyscos:
I think the CI failure is caused by the new ncurses crate requiring a proper `TERM` variable to be set (for example to `xterm`) _during compilation_, since it relies on ncurses itself in the build script to "discover" some constants.

Workarounds include injecting `TERM=xterm` in the CI process, or dropping the ncurses backend entirely.

hrkfdn:
Thanks for the PR ❤️, LGTM! The CD workflows probably also need `TERM`to be adjusted, but we/I can figure this out later.

PR #1494, opened 2024-08-04
Add Seeked signal to Mpris interface

### LINKED ISSUE #1492: Add support for `Seeked` signal on mpris interface
According to the [mpris2 spec](https://specifications.freedesktop.org/mpris-spec/latest/Player_Interface.html#Signal:Seeked), a player should provide the `Seeked` signal to show when the playback position has changed due to seek activity.

I would like support for this signal to be added to ncspot's mpris interface.

n/a

**Additional context**
n/a

Sorry this is short - I hope the request is self-explanatory. Thanks for your work on `ncspot`, it's greatly appreciated.

elParaguayo:
NB I am new to Rust so may have made some horrible mistakes in here!

## Describe your changes

The Mpris2 spec includes a `Seeked` signal which should be fired when the track position changes in an unexpected way i.e. when the user seeks to a different part of the track.

This PR implements this signal on seek events and also when a new track begins. The latter is not strictly required but has been observed in other players (e.g. VLC).

Closes #1492

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Emit MPRIS `Seeked` signal
+
 ### Fixed
 
 - Switch to OAuth2 login mechanism
--- a/src/mpris.rs
+++ b/src/mpris.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::use_self)]
+
 use log::info;
 use std::collections::HashMap;
 use std::error::Error;
@@ -6,6 +8,7 @@ use std::time::Duration;
 use tokio::sync::mpsc;
 use tokio_stream::wrappers::UnboundedReceiverStream;
 use tokio_stream::StreamExt;
+use zbus::object_server::SignalContext;
 use zbus::zvariant::{ObjectPath, Value};
 use zbus::{interface, ConnectionBuilder};
 
@@ -314,6 +317,9 @@ impl MprisPlayer {
         self.queue.get_current().is_some()
     }
 
+    #[zbus(signal)]
+    async fn seeked(context: &SignalContext<'_>, position: &i64) -> zbus::Result<()>;
+
     fn next(&self) {
         self.queue.next(true)
     }
@@ -472,6 +478,8 @@ pub enum MprisCommand {
     EmitVolumeStatus,
     /// Emit metadata
     EmitMetadataStatus,
+    /// Emit seeked position
+    EmitSeekedStatus(i64),
 }
 
 /// An MPRIS server that internally manager a thread which can be sent commands. This is internally
@@ -538,6 +546,10 @@ impl MprisManager {
                 }
                 Some(MprisCommand::EmitMetadataStatus) => {
                     player_iface.metadata_changed(ctx).await?;
+                }
+                Some(MprisCommand::EmitSeekedStatus(pos)) => {
+                    info!("sending MPRIS seeked signal");
+                    MprisPlayer::seeked(ctx, &pos).await?;
                 }
                 None => break,
             }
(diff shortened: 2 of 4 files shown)


hrkfdn:
Thanks :heart:, tested it quickly and it looks pretty good to me! Could you update the Changelog and check on the Clippy errors? After that I'll happily merge it :)

elParaguayo:
Thanks. I've looked at those 2 `cargo clippy` errors and I'm struggling to see how they were caused by my changes.

```
error: unnecessary structure name repetition
  --> src/mpris.rs:78:6
   |
78 | impl MprisPlayer {
   |      ^^^^^^^^^^^ help: use the applicable keyword: `Self`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#use_self
   = note: requested on the command line with `-D clippy::use-self`
```
I don't see how adding a new function caused that. I also don't understand how this gets fixed. The link refers to a `new` function but we don't have that here - is that the problem?

```
error: all variants have the same prefix: `Notify`
   --> src/mpris.rs:470:1
    |
470 | / pub enum MprisCommand {
471 | |     /// Notify about playback status and metadata updates.
472 | |     NotifyPlaybackUpdate,
473 | |     /// Notify about volume updates.
...   |
476 | |     NotifySeekedUpdate(i64),
477 | | }
    | |_^
    |
    = help: remove the prefixes and use full paths to the variants instead of glob imports
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#enum_variant_names
    = note: `-D clippy::enum-variant-names` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::enum_variant_names)]`
```
Again, this should have been an issue before this PR as the two names had the same prefix (and suffix) - but maybe it needs at least 3 members before it fires. To fix this, I'd need to rename existing values. I'm fine to do that if you're ok with it.

ThomasFrans:
That first one is quite simple to solve generally. I think it means you used the literal name `MprisPlayer` somewhere inside an implementation on `MprisPlayer`. In that case it's better to use `Self` to refer to `MprisPlayer`. It helps when searching the codebase, so you don't get 100s of results when looking for an identifier, but only the definition.

For the second one I'd imagine the lint only triggers when you have 3 or more variants with the same prefix, although I'm not sure. I also don't understand what the suggestion means in this case. I've never seen that one either.

elParaguayo:
Thanks for the comments.

> That first one is quite simple to solve generally. I think it means you used the literal name `MprisPlayer` somewhere inside an implementation on `MprisPlayer` (in a function inside the `impl MprisPlayer` block). In that case it's better to use `Self` to refer to `MprisPlayer`. It helps when searching the codebase, so you don't get 100s of results when looking for an identifier, but only the definition.

Unless I'm blind, I haven't done this at all. There an `MprisPlayer` reference in the `impl MprisManager` block but I can't use `Self` there.

> For the second one I'd imagine the lint only triggers when you have 3 or more variants with the same prefix (https://dev-doc.rust-lang.org/nightly/clippy/lint_configuration.html#enum-variant-name-threshold). I also don't understand what the suggestion means in this case. I've never seen that one either.

This is probably just solved by renaming the variants but I'll leave this until I can fix the first one.

haruInDisguise:
Hey, @elParaguayo.  A PR of mine (#1515) has been merged that restructures how MPRIS events are emitted, making this PR (in its current state) incompatible with the main branch.
> Unless I'm blind, I haven't done this at all. There an MprisPlayer reference in the impl MprisManager block but I can't use Self there.

From what I can tell, this is an issue causes by zbus, upon expanding the ```interface(...)``` procedural macro. You could add a clippy exception (```#![allow(clippy::use_self)]```) to the start of ```mpris.rs```.

elParaguayo:
@haruInDisguise thanks for letting me know. I'll rebase this PR and fix as necessary.

hrkfdn (approved):

hrkfdn:
Thank you :) Merged with some very minor changes

elParaguayo:
Thanks for the changes. Beyond my current skill level so appreciate you improving my PR.

PR #1515, opened 2024-09-11
MPRIS: Restructuring and cleanup

### LINKED ISSUE #1514: MPRIS behavior is inconsistent
The mpris "status reports" emitted by ncspot are a bit inconsistent and sometimes redundant.
For example, 'Metadata' and 'PlaybackStatus' reports are one and the same and both being emitted
upon any change to the player's internal state.
The expected behavior would be a clear separation of those events ('Metadata' being sent upon song changes, etc.).

While the current implementation may technically function for most software, it's lack of clearly separate events might prevent software to work properly with ncspot.

I've opened a PR that addresses these issues, in a way that I considered to be "expected" 

**Additional context**
See my PR for additional notes.

haruInDisguise:
## Describe your changes

My pr addresses some of the inconsistencies in ncspot's mpris
implementation. While the previous logic is technically good enough, it
is inflexible and reports redundant information.
I've tried to address those issues to make it easier for software, that processes mpris events in
some way, to accurately react to player updates.

- 'Metadata' and 'Playback' updates have been separated into their own
  command
- Mpris commands are only emitted from within spotify.rs
- Some parts of the application creation logic has been
  restructured to allow for mpris events to be emitted upon startup
    - The initial song loading code has been moved from 'Queue::new'
      into 'Application::new'.

This closes  #1514 

Edit: An example of software that requires strict MPRIS compliance would be the [MPRIS module for yambar](https://codeberg.org/dnkl/yambar/pulls/402) that I am working on.

### DIFF
--- a/src/application.rs
+++ b/src/application.rs
@@ -12,7 +12,7 @@ use signal_hook::{consts::SIGHUP, consts::SIGTERM, iterator::Signals};
 
 use crate::command::Command;
 use crate::commands::CommandManager;
-use crate::config::Config;
+use crate::config::{Config, PlaybackState};
 use crate::events::{Event, EventManager};
 use crate::library::Library;
 use crate::queue::Queue;
@@ -140,6 +140,27 @@ impl Application {
         #[cfg(feature = "mpris")]
         spotify.set_mpris(mpris_manager.clone());
 
+        // Load the last played track into the player
+        let playback_state = configuration.state().playback_state.clone();
+        let queue_state = configuration.state().queuestate.clone();
+
+        if let Some(playable) = queue.get_current() {
+            spotify.load(
+                &playable,
+                playback_state == PlaybackState::Playing,
+                queue_state.track_progress.as_millis() as u32,
+            );
+            spotify.update_track();
+            match playback_state {
+                PlaybackState::Stopped => {
+                    spotify.stop();
+                }
+                PlaybackState::Paused | PlaybackState::Playing | PlaybackState::Default => {
+                    spotify.pause();
+                }
+            }
+        }
+
         #[cfg(unix)]
         let ipc = if let Ok(runtime_directory) = utils::create_runtime_directory() {
             Some(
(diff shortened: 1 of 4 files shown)


src/mpris.rs
```diff
@@ -463,11 +463,14 @@ impl MprisPlayer {
...
-    /// Notify about playback status and metadata updates.
-    NotifyPlaybackUpdate,
-    /// Notify about volume updates.
-    NotifyVolumeUpdate,
+    /// Emit playback status
+    Playback,
+    /// Emit volume
+    Volume,
+    /// Emit metadata
+    Metadata,
```
hrkfdn, line 473: Right now the naming is not very clear as it's neither a command name although the enum type indicates a command-like naming.

Could you either change the enum to something like `MprisUpdateCommand`/`MprisEmitCommand` or have the single fields indicate something similar in the prefix/suffix? WDYT?

src/application.rs
```diff
@@ -140,6 +137,27 @@ impl Application {
         #[cfg(feature = "mpris")]
         spotify.set_mpris(mpris_manager.clone());
 
+        // Load the last played track into the player
```
hrkfdn, line 140: Cool, makes sense moving this here! :+1:

hrkfdn (commented):
Thanks for the PR :heart: I only have a minor nitpicky comment regarding the naming. WDYT?

src/mpris.rs
haruInDisguise: Thanks for the review! This might have been the result of a careless ```sed``` command to fix clippy complaints. Sorry, this should not have slipped through, and was pretty careless of me.
What do you think about ```EmitPlaybackStatus``` etc.? I find that a bit clearer than, ```NotifyPlaybackUpdate``` since basically just send a message into the void and do not "update" anything.

hrkfdn: No worries! I'm fine with either, as you prefer :)

haruInDisguise:
I've chosen my naming convention (:p) and also added an exception so clippy does not complain to us about enums starting with the same prefix.

hrkfdn (approved):

hrkfdn:
Looks good, let's get it in + sorry for the delay! :)

haruInDisguise:
No worries! Thanks for taking the time!

Issue #1528, opened 2024-10-05
Ncspot built with cover feature failed to start


yelaco:
**Describe the bug**
Hi, I tried installing and building (`--release`) ncspot with cover feature. But they both failed to start the program after authentication.

**To Reproduce**
1. Setup
- Install: `cargo install --locked ncspot --features cover`
- Build: `cargo build --release --features cover`
2. Run: 
`ncspot` 
`./target/release/ncspot`

**System (please complete the following information):**
 - OS: Ubuntu 22.04.5 LTS x86_64
 - Terminal: Alacritty
 - Version: 1.1.2
 - Installed from: cargo

**Backtrace/Debug log**
Please attach a debug log and backtrace if ncspot has crashed.

[backtrace.log](https://github.com/user-attachments/files/17326227/backtrace.log)

![image](https://github.com/user-attachments/assets/542e84ff-c110-467c-b623-4641b2e1c4af)

**Additional context**
The installation/build without cover feature works fine
The build with cover feature and no `--release` tag works fine (`cargo build --features cover`)

hrkfdn:
This is the log. The backtrace should be in the cache folder (~/.cache/ncspot) :)

yelaco:
:> here you go.

[backtrace.log](https://github.com/user-attachments/files/17326150/backtrace.log)

moabeat-berlin:
Seems to be incorrectly initialized variables. This is taken from `ui/cover.rs`.

```
        let (rows, cols, mut xpixels, mut ypixels) = unsafe {
            let query: (u16, u16, u16, u16) = (0, 0, 0, 0);
            ioctl(1, TIOCGWINSZ, &query);
            query
        };
```

Since `rows` and `cols` are initialized with 0, it leads to the divide by zero in the backtrace, if they are not changed.

```
        let font_size = Vec2::new((xpixels / cols) as usize, (ypixels / rows) as usize);
```

Maybe simply initializing them with 1 would be just fine?

HurricanePootis:
I made a pull request that should hopefully fix this

hrkfdn:
Fixed with #1542 + released via ncspot 1.2.1

PR #1533, opened 2024-10-16
fix(cover): init variables with value of 1

HurricanePootis:
## Describe your changes
Initializes `rows` and `cols` with a value of 1 instead of 0 to avoid dividing by zero.
https://github.com/hrkfdn/ncspot/issues/1528
- [X] Changelog was updated with relevant user-facing changes (eg. not dependency updates,

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [Unreleased]
+
+### Fixed
+
+- Initialize variables used in cover with value 1 to avoid divide by zero error
+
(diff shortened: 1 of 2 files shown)


hrkfdn (approved):
Thanks for the PR!

hrkfdn:
Could you update the Change log?

HurricanePootis:
I've never used a change log before. Isn't that only whenever there is a new version to be released?

HurricanePootis:
@hrkfdn 

Changelog was updated

arielcrivas:
I just want to report that with this patch, ncspot runs fine but crashes when I try to change to the cover tab with F8
Same backtrace
[backtrace.log](https://github.com/user-attachments/files/17467956/backtrace.log)
Thx in advance

antonnyst:
After looking into this bug and patch I think I have found out the root cause, the unsafe block expression fails to return the values, I was able to fix this by adding `mut` to the `query`. 

I created a new PR for this fix: #1542

hrkfdn:
Does that mean this one can be closed?

PR #1542, opened 2024-10-29
fix(cover): Add mut to fix incorrect get of windowsize in cover

antonnyst:
## Describe your changes
Something seems to have broken the return of the variables from the unsafe block, leading to division by zero.
By making the query variable mutable the return works. 

#1528


hrkfdn (approved):
Makes sense, thanks!

PR #1573, opened 2024-12-30
docs: remove remaining password manager documentation

ThomasFrans:
## Describe your changes
Remove remaining password manager documentation and add some notes about the new OAuth2 flow.

\/


ThomasFrans:
I guess "Automatic authentication using a password manager" is technically still correct if you are using a browser extension :thinking:

hrkfdn:
I think that's rather on the user end setup, so it makes sense to remove it.

PR #1601, opened 2025-03-06
Update edition 2024

ThomasFrans:
## Describe your changes

Mostly followed the official migration guide, and did more builds/tests/... to make sure everything works fine after every step.

1. Updated all dependencies
  - Ran `cargo update`
  - Verified all code still built fine by running various version of `cargo build` (with all features, without default features, for all targets, various profiles, all workspaces...)
  - Verified all tests still ran fine
2. Update the Rust edition
  - Run `cargo fix --edtion` (for all features, all targets, for all workspaces...)
  - Read through the warnings. The only generated warnings were about [the new drop order of temporary variables](https://doc.rust-lang.org/edition-guide/rust-2024/temporary-tail-expr-scope.html) for variables with a custom implementation of `Drop`. The warnings mention that the new drop order is rarely a problem. Didn't change any of the code the warnings mentioned (later tests seem to suggest that everything works fine).
  - Went through the diffs. There were two more functional changes: use of the new `expr_2021` and `unsafe` for `env::set_var`.
  - `if let ... else ...` is also changed in a lot of places to `match` statements. Doesn't seem to change functionality.
  - Updated `Cargo.toml` files to share common options across workspace members (`edition`, `authors`, `license` and `repository`). This ensures these options remain the same across workspace members during future edition updates.
  - Verified all code still built fine by running various version of `cargo build` (with all features, without default features, for all targets, various profiles, all workspaces...)
  - Verified all tests still ran fine
  - Verified `ncspot` still ran fine by logging in, navigating the UI and playing a song.
3. Format all code with `cargo fmt --all`

\/


ThomasFrans:
Just wanted to try a Rust edition update to see how difficult or easy it is. I made a PR in case you'd like to accept this change since I now had the result of the update anyway. It was mostly for exploring this process myself, so feel free to close if you'd like to do the edition update yourself or at a later point in time :slightly_smiling_face:

hrkfdn:
Awesome work, thank you! ❤️  I'll try to review this in the coming days.

src/commands.rs
```diff
@@ -332,33 +332,40 @@ impl CommandManager {
...
+                    CommandResult::View(view) => {
+                        s.call_on_name("main", move |v: &mut Layout| {
+                            v.push_view(view);
+                        });
+
+                        Ok(None)
+                    }
+                    _ => self.handle_default_commands(s, cmd),
+                },
+            },
```
hrkfdn, line 368: Was this suggested by the toolchain? I find the nested matches harder to read than the `if let` expressions.

hrkfdn (commented):
Thanks! It looks good to me, I haven't read about the changes in 2024 yet, but is `if let` deprecated over match?

src/commands.rs
ThomasFrans: Yeah, I also find it way harder to read. There were even places with like 8 closing curly brackets to close all the nested scopes. This was automatically done by Cargo when I executed `cargo fix --edition`. I have no clue why.

The only manual changes I made in this PR were changes to the workspace definitions in the two `Cargo.toml` files to make future edition updates easier. I think those are in the second commit where I also increased the edition number from 2021 to 2024.

ThomasFrans:
> Thanks! It looks good to me, I haven't read about the changes in 2024 yet, but is `if let` deprecated over match?

After quickly looking it up, I think it's because of [this change](https://doc.rust-lang.org/edition-guide/rust-2024/temporary-if-let-scope.html#migration). This page also mentions the preference for `match` statements.

hrkfdn:
> > Thanks! It looks good to me, I haven't read about the changes in 2024 yet, but is `if let` deprecated over match?
> 
> After quickly looking it up, I think it's because of [this change](https://doc.rust-lang.org/edition-guide/rust-2024/temporary-if-let-scope.html#migration). This page also mentions the preference for `match` statements. The page also mentions that you can manually decide whether you want to revert the change to match statements. I didn't see that during the migration as I only followed the general guide, which never mentioned this.
> 
> Want me to change the match statements back to `if let`? I would need to look what the new behavior of `if let` is in edition 2024 to ensure that new behavior is compatible with the old code.

I think generally it's fine, only for the nested ones it reallly hinders readability. I'd say those should be reverted, WDYT?

ThomasFrans:
I reverted most of the match statements back to their original state. I built and tested and the basic stuff seems to work fine. Playback, navigation, mouse interaction, context menus, opening stuff... The new behavior of `if let` probably wasn't problematic for all the cases where it was changed to `match`.

I just don't know how to resolve issues with `Cargo.lock`. I've never had to deal with merge conflicts in lock files and I can't imagine you have to manually resolve them. Do you know how to solve them? @hrkfdn

hrkfdn:
I will take care of the lockfile and then merge. Thanks a lot! :)

hrkfdn (approved):

Issue #511, opened 2021-05-03
Make play/pause indicators configurable


fwcd:
The player currently displays the 'play icon' when a song is playing and the 'pause icon' when it's not playing. Since most traditional audio player GUIs do this the other way around, is there a specific reason for the current way?

hrkfdn:
Hmm, I'm not sure about this one. It's more of a playback status indicator than a button, that's why it's the way it currently is. I think if it were a button to toggle playback, it would make sense to reverse the icons, but in this case it seems rather counterintuitive to me. Any other thoughts?

JojiiOfficial:
I personally see it as button, even though I use 'P' to toggle playback. As @fwcd already pointed out, most audio player GUIs treat it as button and I think this point of view is preserved while using ncspot (at least in my case). In addition you can click the icon with the mouse to toggle playback as well. I was kinda confused the first time I used ncspot why the indicator status was inverted.

codicodi:
It seems unintuitive to me as well (especially since I also use a custom MPRIS widget that always shows the opposite of what ncspot does). Maybe icons could be made configurable, similar to how colors are?

ClashTheBunny:
I do have to look at the time tick when I don't hear anything because it's not intuitive to me.

AndydeCleyre:
Ha, I came here looking for the opposite bug. 

I had no idea that was a button, so was sure it was a bug, showing the pause icon while the state is playing, as I considered the widget to be just like the shuffle icon, which shows a shuffle symbol when the state is shuffle.

So now I try clicking the shuffle icon (and elsewhere), and realize that the whole status bar is one big play/pause button.

So that little indicator, to me, is *not* a button, but an indicator, that happens to reside within a large toggle button.

I would appreciate an option to have the indicator show the current state again, rather than the opposite of the current state.

BachoSeven:
https://xkcd.com/1172/ `/s`

hrkfdn:
Haha, I feared this would happen 😄 It was meant to be an indicator actually. Maybe making it configurable is a good idea.

hrkfdn:
Okay, I have reverted to the old behaviour and made this a configurable setting.

fwcd:
While I would argue that defaulting to the flipped version would be more intuitive for people used to traditional, button-style UIs, having it configurable is definitely nice and should make everyone happy :+1:

hrkfdn:
[...]

PR #1620, opened 2025-04-10
fix: indicators were wrong

nettproxy:
I fixed the Indicators, as it was showing "Playing" when the track was paused, and "Paused" when the track was playing.

### DIFF
--- a/src/ui/statusbar.rs
+++ b/src/ui/statusbar.rs
@@ -58,8 +58,8 @@ impl StatusBar {
         };
 
         match status {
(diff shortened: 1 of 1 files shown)


hrkfdn:
Hey! Thanks for the PR.

This seems to be somewhat controversial, people have different interpretations for which indicator should be shown. See also this discussion https://github.com/hrkfdn/ncspot/issues/511.

Back when we therefore decided to make it configurable via `flip_status_indicators`.

PR #1621, opened 2025-04-13
chore: remove unused `rustfmt.toml` file

ThomasFrans:
This configuration file is empty and therefore doesn't do anything. It makes sense to remove it to keep the files and directories in the root of the project as minimal as possible. This makes it easier for new contributors to get a quick overview of the project. The more files there are in the root of a project, the harder it gets to tell what's important.

## Describe your changes

- Remove the empty `rustfmt.toml` configuration file for `rustfmt` from the root directory of the project

\/


ThomasFrans:
Been wanting to make a PR for this for a while, but it's such a small change that it's almost not worth it :smile:

hrkfdn:
Thank you :)

Issue #1631, opened 2025-05-09
Application becomes unresponsive when attempting to play unavailable tracks on repeat


joehasson:
**Describe the bug**
 The application becomes unresponsive under certain circumstances when I try to play an unavailable track. This seems to occur when my queue consists entirely of unplayable tracks and the playback mode is either repeat playlist (R) or repeat track (R1).  CPU and memory utilisation of the app also grows to exceed the levels seen during normal usage.

**To Reproduce**
1. Open the application and press `r` to set the playback mode to either R or R1.
2. Fill the queue entirely with unavailable tracks eg by playing this playlist https://open.spotify.com/playlist/5jlJlLvxb6ncAD1eRLvaU2
3. The application become unresponsive while system resource usage spikes (observable e.g. with htop).

**Expected behavior**
Application should not become unresponsive. Ideally it might handle the attempt by the user to play an unavailable track gracefully, for example flashing an error message and continuing to play the currently playing track. 

**System (please complete the following information):**
 - OS: Linux Mint 21.1
 - Terminal: Alacritty
 - Version: 1.2.2
 - Installed from: cargo build (debug or release) from main branch. 

Notably I can't reproduce this with the latest version of ncspot installed via cargo  (1.2.2), so this likely has been introduced since that release was made. 

**Backtrace/Debug log**
Running `tail -f` on the log file I observe a looping pattern like this
```
[2025-05-09][23:09:25] [ncspot::application] [TRACE] event received: FinishedTrack
[2025-05-09][23:09:25] [ncspot::spotify] [DEBUG] Sending mpris command: EmitPlaybackStatus
[...]

PR #1634, opened 2025-05-12
Prevent event loop starvation by unplayable tracks

joehasson:
## Describe your changes
Fixes the event loop starvation problem identified in #1631 by adding checks in the `Queue::next` method which determine whether there are any playable tracks remaining given the current queue contents and repeat settings. If there are no playable tracks left, playback is stopped, preventing the infinite loop which caused the application to stall. 

This issue was introduced in commit e3e75f7, which added skipping of unplayable tacks, but didn't account for the repeat mode edge cases. These changes maintain the skipping behaviour while accounting for those edge cases.

### Alternatives considered
-  Add an invariant that  `queue::Queue`  items are playable. I initially implemented a solution enforcing this invariant at compile time using the newtype pattern in [this commit](https://github.com/hrkfdn/ncspot/compare/main...joehasson:ncspot:fix/unplayable_tracks_in_queue_newtype_solution). On the one hand this seems to provide a stronger guarantee that this or similar issues won't re-emerge in the future. However the changes required by this approach are larger and wider in scope so I have opted for the more contained solution in this PR as it addresses the issue with minimal changes and no change to the existing `Queue` API. I'd be glad to raise a PR for the more comprehensive type-level solution if that seems preferable.
- [x] Leaving changelog untouched as this issue only affects unreleased versions

### DIFF
--- a/src/queue.rs
+++ b/src/queue.rs
@@ -13,6 +13,7 @@ use crate::library::Library;
 use crate::model::playable::Playable;
 use crate::spotify::PlayerEvent;
 use crate::spotify::Spotify;
+use crate::traits::ListItem;
 
 /// Repeat behavior for the [Queue].
 #[derive(Display, Clone, Copy, PartialEq, Eq, Debug, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


joehasson:
Hi - I think it is best to keep these changes in the `next` method, since this is already where we are determining what (if anything) to play based on the queue contents and repeat settings. Putting these new checks elsewhere would fragment this logic. Does that reasoning make sense to you?

hrkfdn:
OK by me. I was wondering if we could reduce the redundancy and move it to a more central place, but your point is valid. Merged! :)

hrkfdn (approved):

Issue #1638, opened 2025-05-16
UI freezes in queue tab if mouse click used to play


Fatcat560:
**Describe the bug**
Trying to play a song in the queue by "double clicking" on

**To Reproduce**

**Expected behavior**

**Screenshots**

**System (please complete the following information):**
 - OS: [e.g. Linux, macOS]
 - Terminal: [e.g. GNOME Terminal, Alacritty]
 - Version: [e.g. 0.4.0, `master` branch]
 - Installed from: [e.g. AUR, brew, cargo]

**Backtrace/Debug log**
Please attach a debug log and backtrace if ncspot has crashed.

Instructions on how to capture debug logs: https://github.com/hrkfdn/ncspot#debugging

For backtraces, make sure you run a debug build of ncspot, e.g. by running the
command mentioned in the [compilation
instructions](https://github.com/hrkfdn/ncspot#compiling).  You can find the
latest backtrace at `~/.cache/ncspot/backtrace.log`.

**Additional context**

hrkfdn:
Thanks for the report, I can reproduce it. I didn't have time to fix it yet but did some digging:

I think the problem lies here: https://github.com/hrkfdn/ncspot/blob/c294857002f4237085414a402e59e936ca0d1d2f/src/ui/listview.rs#L391-L399

This is the logic for a double click on an item. It will clear the queue and then append all the items shown in this view. This works fine for non-queue views (i.e. albums), but inside the queue view this causes problems.

That logic is also duplicated for the play command (i.e. when a user presses return on an item): https://github.com/hrkfdn/ncspot/blob/c294857002f4237085414a402e59e936ca0d1d2f/src/ui/listview.rs#L486-L494

For the queue view (which uses `ListView` under the hood), the command behavior is overwritten to only play the selected item without clearing the queue: https://github.com/hrkfdn/ncspot/blob/c294857002f4237085414a402e59e936ca0d1d2f/src/ui/queue.rs#L121-L124
I think we need the same behavior for the mouse click. Sharing this here as notes for me or someone else to fix it.

Fatcat560:
@hrkfdn thanks for reproducing! I'm not very knowledgeable about the code base but if I find the time, I'd love to help fix this issue. Looking at the pieces of code you provided makes me think this should be an easy-ish fix so maybe the perfect opportunity to try my luck.

Fatcat560:
Update: I have a working version after some tinkering. I'll see if I can somehow clean it up a bit before opening a PR though.

PR #1642, opened 2025-05-21
Prevent Queue UI from freezing when clicking on a selected item

Fatcat560:
## Describe your changes

In order to enable the queue view to play a title by clicking on the selected row, I did the following changes:

- Most of the `on_event`  logic moved into a separate public function in order to reuse the existing code as much as possible in the queue ui.
- This new `handle_mouse_event`  function either directly handles the incoming event or, in the case of the play option, returns a `Command` which then can be handled separately by different views. This in theory could enable more commands in the future if required.

[#1638](https://github.com/hrkfdn/ncspot/issues/1638)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Skip unplayable tracks
+- Queue UI correctly plays a track when clicking on an already selected item
 
 ## [1.2.2]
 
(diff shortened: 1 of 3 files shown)


hrkfdn (approved):
Thanks so much for your contribution :heart:

I've been breaking my brain pondering whether there's a more generic way to implement this, i.e. should we have playable item click events propagate instead of the `ListView` triggering a `Play` command, etc., because this feels a little specific. But I can't come up with anything and the `ListView` is not very generic anyway, so I'd merge this in a bit, especially because it fixes a rather nasty bug :)

Fatcat560:
@hrkfdn And thank you for maintaining this app :100:  I've been very happy using it and I'm glad I was able to give back to the project in some way.

I didn't want to touch too many things with this bug fix as my first contribution, maybe I'll invest more time reading the code to better understand the inner workings. That way I may be able to contribute once more in the future.

hrkfdn:
Yeah I understand, I'm also a bit rusty with the codebase 😅 Thanks again!

PR #1675, opened 2025-08-04
Add playlist `reverse` command

floscr:
## Describe your changes

Implements a reverse playlist command to reverse the order of tracks in a playlist.
Reverse was not implemented as a sort method, as I don't see it as a dynamic order, but as a do once command.
When reversing playlist we also clear the current sort order.

Adds it under the default keybinding `Shift+R`

N/A

- [x] Changelog was updated with relevant user-facing changes (eg. not dependency updates, not performance improvements, etc.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Playlist `reverse` command with keybinding `Shift+R`
+### Added
 
 ## [1.3.0]
 


hrkfdn:
Hi there, apologies for the late review. Generally I think this is a nice idea, but I wonder if this should be applicable to queues instead, which are more like a scrap playlist that can then be saved?

Because the way this is implemented right now will reverse playlists but there's no way to save them from the playlist view, right?

floscr:
Hey no problem!

My use case was to quickly flip the order of items in a playlist where I add stuff and sort it over time.
This wouldn't have quite fit with the `sort` command, as sorting by `added` doesn't do the same once you re-arrange items in a playlist.
(In spotify this is called custom order and there's no quick way to reverse this list)

This command could possibly be extended to any view IMO, 
but maybe a custom dynamic sort order would be a better fit (So reverse of no sort) and then having a quick way to toggle this.

But I do like the UX of having just a ":reverse" command.

PR #1692, opened 2025-08-28
feat: Use Spirc for player logic

### LINKED ISSUE #83: Missing spotify-connect support
Ncspot doesn't seem to implement spotify-connect support meaning that it doesn't appear in other players (official app, website) as a device. Consequently, it is not possible to control remote device, that is playing music with my local ncspot.

It shouldn't be that hard to implement because [spotifyd](https://github.com/Spotifyd/spotifyd) knows it and it is also based on the [librespot](https://github.com/librespot-org/librespot)

hrkfdn:
**Heavily WIP, only does basic playback so far.**

## Todo
- [ ] Update player context via [LoadRequest](https://docs.rs/librespot-connect/0.7.0/librespot_connect/struct.LoadRequest.html)
- [ ] Handle `TrackChanged` events
- [X] Handle `VolumeChanged` events
- [ ] Implement debouncing for volume updates to prevent rate limit exhaustion

## Describe your changes
The idea is to use librespot `Spirc` instead of handling the player logic within ncspot. This should reduce complexity a bit, but also allows us to integrate with Spotify Connect, which in turn would enable ncspot to be controlled via Spotify Apps.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2127,6 +2127,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "librespot-connect"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b2a21e9e19b906c6676b1a515fc5529d5414f69049f5a167dfb2ea3206e5a27"
+dependencies = [
+ "futures-util",
+ "librespot-core",
+ "librespot-playback",
+ "librespot-protocol",
+ "log",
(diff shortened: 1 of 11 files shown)


Bettehem:
If implemented, I would hope for the spotify-connect feature to be an optional toggle, as the way it currently works is perfect for me as i can have multiple instances of ncspot running but with spotify-connect that is not possible

PR #1702, opened 2025-09-09
removed ConsoleOnly from Caregories

mrsteve0924:
Including ConsoleOnly prevents application finder from showing desktop file

### DIFF
--- a/misc/ncspot.desktop
+++ b/misc/ncspot.desktop
@@ -6,5 +6,5 @@ TryExec=ncspot
 Exec=ncspot
 Icon=ncspot
 Terminal=true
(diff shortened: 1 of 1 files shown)


hrkfdn:
Hey, according to the [specs](https://specifications.freedesktop.org/menu-spec/latest/additional-category-registry.html) `ConsoleOnly` sounds correct. Which application finder is this and why is it hiding console applications?

mrsteve0924:
i am using wayfire's default application menu included in their panel (wf-panel).

I added the ConsoleOnly back to the Categories section and it seems to function properly now.  Strange.  

I guess we can close this PR.

thanks for the review.

PR #1720, opened 2025-10-10
fix: prevent crash when song has no artist

ThomasFrans:
When loading this playlist
(https://open.spotify.com/playlist/2onE4ObADgp1NuBzciYF0Q), ncspot would crash. The exact cause is not clear, but it seems to be caused by the songs not being available and therefore not having a valid artist. Therefore ensure we have a default value for artists in case one isn't available.

## Describe your changes
- Prevent a crash when loading certain playlists that contain unavailable songs (e.g., [this one](https://open.spotify.com/playlist/2onE4ObADgp1NuBzciYF0Q) in Belgium)

\/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Playlist retrieval crashing when list contains podcast episodes
 - Crash when shifting a song by an amount greater than the queue's length
+- Crash when displaying songs that do not have an (available) artist
 
 ## [1.3.1]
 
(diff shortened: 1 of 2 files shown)


hrkfdn:
Huh, interesting. But the song is playable?

ThomasFrans:
The songs in that playlist aren't playable, at least not in Belgium. The error I got was an unwrap on the artist field of those songs, which for some reason does not contain any artists.

hrkfdn:
I see, so the player will just skip them?

ThomasFrans:
I think so. I didn't try on this album as all the songs are unavailable, but in other playlists with only a few unavailable songs, ncspot just skips them if I remember correctly.

hrkfdn:
Let's get it in. Thanks!

PR #1746, opened 2025-12-06
Feat/more vim cmds

nickscip:
## Describe your changes
Added vim motions gg (double-press 'g' to go to top) and G (Shift+g to go to bottom) with double-press detection, and updated the user documentation.

- [ x ] Documentation was updated (i.e. due to changes in keybindings, commands, etc.)
- [ x ] Changelog was updated with relevant user-facing changes (eg. not dependency updates,

### DIFF
--- a/src/commands.rs
+++ b/src/commands.rs
@@ -567,7 +567,7 @@ impl CommandManager {
 
         kb.insert(
             "g".into(),
-            vect![Command::Move(MoveMode::Up, MoveAmount::Extreme)],
+            vec![Command::Move(MoveMode::Up, MoveAmount::Extreme)],
         );
         kb.insert(
             "Shift+g".into(),


ThomasFrans:
I've also been missing this, as using Home/End is annoying. Some programs that do not use modes use a single g to go to the top. I wonder whether that would be better as it would reduce the complexity a bit and pressing g twice will sort of work as well.

Bettehem:
I like the idea of having more vim-like navigation too and have been able to achieve this g/G functionality by adding this to my config.toml (as well as some ctrl+d/ctrl+u like binds)
```
[keybindings]
"g" = "move top"
"Shift+g" = "move bottom"
"Ctrl+d" = "move down 10"
"Ctrl+u" = "move up 10"
```

But having such functionality by default could be nice too

hrkfdn:
Thanks @nickscip for the PR, appreciate it. Would introducing default bindings for `g` and `Shift+g` (and the others) as suggested by @Bettehem  also work? I'm not a Vim power user, so I can't say how critical this is and I realize it's not exactly the same behavior, but it would reduce the complexity a bit.

PR #1783, opened 2026-02-11
Fix crash when adding a song to a playlist

AnAngryRaven:
## Describe your changes

Set the initial fetch for the WebAPI token as blocking, as otherwise it resulted in a race condition on application startup where the `user_id` field in `library.rs` could not be assigned in at least some circumstances. This would then ultimately trigger the unwrap for `current_user_id` (line 62 of `contextmenu.rs`), as it depends upon the aforementioned field.

- [ ] Documentation was updated (i.e. due to changes in keybindings, commands, etc.) -- N/A


hrkfdn:
Merged, thank you!

PR #1841, opened 2026-06-27
fix(mpris): prevent volume update before instance created

ThomasFrans:
We now try to update the MPRIS volume before an MPRIS instance exists, which show an error message (with the DEBUG priority for some reason) in the logs. This just ensures we don't do that, preventing that message.

## Describe your changes

- Don't send an MPRIS volume update when initializing volume in the `Spotify` struct.

\/


ThomasFrans:
Noticed this small error while looking into another bug that prevents songs from playing. Saw that I introduced this bug, so felt like I should quickly fix it.

Didn't change the changelog as this isn't user-facing.

hrkfdn (approved):
Thank you, looks reasonable

PR #1863, opened 2026-08-22
feat: vim-style command completion

koloiyolo:
## Describe your changes

Added basic vim-style command completion on `Tab`.

https://github.com/user-attachments/assets/87e0372b-7634-46d3-a0d6-a2ec293b63b1

Refactored `command.rs` module:

* Moved literal `&str` values from [command::Command::basename] to `const` values.
* Reused the `const` values for [command::parse] implementation

N/A 

 [command::Command::basename]: https://github.com/hrkfdn/ncspot/blob/1a36310f71e1b84ed72b1257d8b56d7d3e6cb716/src/command.rs#L232
 [command::parse]: https://github.com/hrkfdn/ncspot/blob/1a36310f71e1b84ed72b1257d8b56d7d3e6cb716/src/command.rs#L362

### DIFF
--- a/src/command.rs
+++ b/src/command.rs
@@ -6,6 +6,97 @@ use std::sync::OnceLock;
 
 use strum_macros::Display;
 
+const QUIT_CMD: &str = "quit";
+const TOGGLE_PLAY_CMD: &str = "playpause";
+const STOP_CMD: &str = "stop";
+const PREVIOUS_CMD: &str = "previous";
+const NEXT_CMD: &str = "next";
+const CLEAR_CMD: &str = "clear";
+const QUEUE_CMD: &str = "queue";
+const PLAY_NEXT_CMD: &str = "playnext";
+const PLAY: &str = "play";
+const UPDATE_LIBRARY_CMD: &str = "update";
+const SAVE_CMD: &str = "save";
+const SAVE_CURRENT_CMD: &str = "save current";
+const SAVE_QUEUE_CMD: &str = "save queue";
+const ADD: &str = "add";
+const ADD_CURRENT_CMD: &str = "add current";
+const DELETE_CMD: &str = "delete";
+const FOCUS: &str = "focus";
+const SEEK: &str = "seek";
+const VOLUME_UP_CMD: &str = "volup";
+const VOLUME_DOWN_CMD: &str = "voldown";
+const REPEAT_CMD: &str = "repeat";
+const SHUFFLE_CMD: &str = "shuffle";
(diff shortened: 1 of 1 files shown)


src/ui/layout.rs
```diff
@@ -282,6 +286,29 @@ impl Layout {
...
         }
     }
+
+    /// Completes the command line when the `cmdline` input uniquely matches a command.
+    ///
+    /// TODO: Add support for aliases.
+    ///
+    /// # Returns
+    ///
+    /// [EventResult::Consumed] when the command match was found, [EventResult::Ignored] otherwise.
```
koloiyolo:
```suggestion
    /// [EventResult::Consumed].
```

I' ll fix this after the review as the CI checks are passed atm.

src/command.rs
```diff
@@ -6,6 +6,98 @@ use std::sync::OnceLock;
...
+const INSERT_CMD: &str = "insert";
+const NEW_PLAYLIST_CMD: &str = "newplaylist";
+const SORT_CMD: &str = "sort";
+const LOGOUT_CMD: &str = "logout";
+const SHOW_RECOMMENDATIONS_CMD: &str = "similar";
+const REDRAW_CMD: &str = "redraw";
+const EXECUTE_CMD: &str = "exec";
+const RECONNECT_CMD: &str = "reconnect";
+
+pub const ALL_COMMANDS: &[&str] = &[
```
hrkfdn, line 54: I haven't tested this, but is this something where we could utilize `strum` maybe, i.e. [AsRefStr](https://docs.rs/strum/latest/strum/derive.AsRefStr.html).

Would be nice if we don't have to maintain a list of all commands and could keep the serialization close to the enum definition.

hrkfdn:
Thanks for the PR! Nice improvement. I had one idea that could possibly simplify this, do you think it's feasible?

src/command.rs
koloiyolo: Hell yeah. Great suggestion. I'll reimplement it this way as soon as I can!

koloiyolo: I checked it. It is generally good approach that would also require [`strum_macros::EnumIter`](https://docs.rs/strum_macros/latest/strum_macros/derive.EnumIter.html) . The issue is that `EnumIter` itself requires the [`std::Default`](https://doc.rust-lang.org/std/default/trait.Default.html) implementation for the internal fields.

Would you like me to implement these also. Honestly for me this doesn't seem like a desired behavior for fields that should be explicitly defined by design.

